PR #46, opened 2015-01-12
Support dockerization

adamc00:
In order to utilise MailDev in a docker container it was useful to remove the restriction on running the HTTP and SMTP services on ports below 1000 since that allows other dockerized services to find them without alteration to their configuration (this is obviously most useful for SMTP). This PR also includes a Dockerfile suitable for producing a working docker image. Happy to make any changes to the structure of this pull request or the method of allowing lower ports if that would suit better. Thanks!


djfarrelly:
Nice, this is great @adamc00! I'm definitely cool with removing that port < 1000 error. It's cleaner to allow it for root users.

PR #50, opened 2015-02-02
fix typo and jslint error

henryqdineen:


djfarrelly:
Thanks @henryqdineen!

PR #49, opened 2015-01-28
View email source. Closes djfarrelly/MailDev#10.

codecustodian:
Thanks for this great project. This adds a display option for the source file.

 selection_001

PR #52, opened 2015-02-24
Allow running multiple instances of MailDev

rpaterson:


rpaterson:
For reasons I won't bore you with I want to run multiple instances of MailDev on the same machine.  So long as I specify different `--smtp` and `--web` ports it works fine - except they both try to write to `/tmp/maildev` and one errors.  This fixes that by appending the process ID to the temp dir path - e.g.: `/tmp/maildev/12345`.  It's a little hacky but seems fine for this purpose.

djfarrelly:
Nice @rpaterson! You could never bore me with ways you're using MailDev :+1: 

I need to get a 0.8.0 release out after this and #49. Hoping to squeeze in some time this week to get that tested and published. As always, thanks for your contribution Roy!

rpaterson:
Oops this introduced a bug - it doesn't work if the /tmp/maildev dir doesn't already exist.  Fix on the way...

PR #53, opened 2015-02-26
Create /tmp/maildev if it doesn't already exist

rpaterson:
Fixes bug introduced in #52


djfarrelly:
Thanks for the quick fix - going to get out 0.8.1 ASAP

PR #55, opened 2015-03-26
added unread items count in the <title>

AlexDpy:
First, thank you for this really nice package !

I just added the count of unread items in the title, so now we can see, f.e., (5) MailDev.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -1,18 +1,18 @@
 <!doctype html>
-<html>
+<html ng-app="mailDevApp" ng-controller="MainCtrl">
     <head>
         <meta charset="utf-8"/>
         <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1"/>
-        <title>MailDev</title>
+        <title>{{ unreadItems > 0 ? '(' + unreadItems + ') ' : '' }}MailDev</title>
         <meta name="description" content="SMTP Server + Web Interface for viewing and testing emails during development.">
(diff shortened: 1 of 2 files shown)


ProPheT777:
:+1:

djfarrelly:
Thanks for this @AlexDpy! Looks good to me, I'll make sure to test and merge it as soon as soon as I can :+1:

djfarrelly:
Hey @AlexDpy! I pulled this down and tested the code out and there are a couple bugs with the current implementation. Using `$scope.unreadItems--;` in the `$scope.markRead` method can actually make the unread count become negative. If a user clicks an already read email, it will decrease the unread count regardless of if it's already been read. 

One thought I had was to recalculate the absolute value each time instead of incrementing/decrementing a counter. I was thinking a function like this:
```javascript
var countUnread = function() {
  $scope.unreadItems = $scope.items.filter(function(email) {
    return !email.read;
  }).length;
};
```
And called that every time we received new mail/read an email! What do you think?

AlexDpy:
You're right. I've updated the PR. I'm just not sure about the usage of the $promise in the loadData(), but it works.
```js
var loadData = function() {
$scope.items = Email.query();
+ $scope.items.$promise.then(function() {
+ countUnread();
+ });
};
```

djfarrelly:
:+1: Thanks @AlexDpy, I'll get this out in `0.9.1`

PR #57, opened 2015-04-10
Add missing -p to docker command

ptt-homme:


djfarrelly:
Thanks for this @ptt-homme!

Issue #47, opened 2015-01-22
Add optional authentication for web GUI


pkat:
It would be great if we could have optional basic auth implemented so that access could be restricted on dev servers for example.

djfarrelly:
Hey @pkat! That's a cool idea, I could see it coming in handy, but it might be best suited for an external package/plugin in the future. What about trying an `.htaccess` file? Would that work for your use case?

pkat:
@djfarrelly nope that won't work, since maildev doesn't run off of apache, its a node app (unless you proxy it). what about adding the node basic-auth package? easy to implement a simple 2 liner basic on a config option. https://www.npmjs.com/package/basic-auth

djfarrelly:
I'm still not quite sure if this should be in the core project. I'm debating if it's outside of most users use cases. There are a couple higher priority items that I may work on in the near future, so I most likely won't hack on this, but I'm open to a pull request, sorry!

djfarrelly:
Hey @pkat! A pull request was submitted and merged in today to add this feature. If you would like to use it today you can clone the project, or wait until I release version `0.10.0` later this week which will include that feature!

PR #59, opened 2015-04-13
Adding HTTP auth options

Seidr:
Added HTTP auth option, following from a requirement internally, and Issue #47 

I'm not sure if the way I've done this is correct or not, but it appears to work for our setup. I've had to override the 'listen' export originally provided by 'server', as the static route was allowing bypass of the auth middleware.

If you can think of improvements to be made to this, I'm all ears. Relatively new to NodeJS (though not JS), so completely open to criticism.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,8 @@ If you want to use MailDev with [Docker](https://www.docker.com/), you can use t
       --outgoing-user <user>  SMTP user for outgoing emails
       --outgoing-pass <pass>  SMTP password for outgoing emails
       --outgoing-secure       Use SMTP SSL for outgoing emails
+      --web-user <user>       HTTP basic auth username
+      --web-pass <pass>       HTTP basic auth password
       -o, --open              Open the Web GUI after startup
       -v, --verbose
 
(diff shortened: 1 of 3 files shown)


Seidr:
While this works, after doing a bit of reading on middleware in Node, I'm guessing that this provides inadequate protection to all routes. I'll revisit this, with a more complete solution. Do not merge for the time being - I'll reopen when I'm happy with it.

Seidr:
In order to correctly implement Basic Auth, I've changed how the web lib sets up the routes. Essentially, this will ensure that the Basic Auth middleware is added prior to all routes. Due to requiring exports of its own (instead of purely exports from 'server'), I've implemented an intermediary for the 'server.listen' method. All routes are now correctly protected, instead of just the '/' route.

Again, any feedback is welcome.

djfarrelly:
Hey James (@Seidr)! I just tested this out and this is a great solution for adding basic http auth. I may do a little bit of rearranging on this one so there's less logic in `index.js` and may refactor `web.js`. I like how adding of the routes is separated out. On other Node projects I often use a similar approach of passing the `app` object to a function like `addRoutes`.

Issue #60, opened 2015-04-15
Deleting mails is slow - the entire screen redraws.


Grynn:
It's kind of hard to describe, but the whole screen refreshes whenever I delete an email. This makes deleting slow. 

I guess it's not much of priority, but I left MailDev running for a few weeks as an open SMTP server (which does not relay anywhere, so it's not contributing to spam) and got hundreds of crappy emails that I have to delete. It's annoying that the refresh takes so long.

Would be very happy to contribute, if someone could give me a few pointers - I don't know angular so there's that learning curve too.

Small video of the whole refresh thing:
https://dl.dropboxusercontent.com/u/9470118/Recording%20%232.mp4

Grynn:
https://github.com/djfarrelly/MailDev/issues/61 is a better feature request and has code attached!

PR #61, opened 2015-04-16
Prevent full refresh of UI when deleting a single email

Seidr:
This pull request addresses issue #60. Instead of performing a full refresh and redrawing the whole UI, this will, upon a successful delete, first of all find the email element for the email being deleted, navigate to the root path if the email is currently selected, and then remove the email element from its parent.

### DIFF
--- a/app/scripts/controllers/item.js
+++ b/app/scripts/controllers/item.js
@@ -117,9 +117,8 @@ app.controller('ItemCtrl', [
     $scope.delete = function(item) {
 
       Email.delete({ id: item.id }, function(email) {
-        var emailElement = document.getElementsByClassName("email-"+item.id)[0];
-        if (emailElement.classList.contains("current")) $location.path("/");
-        emailElement.parentNode.removeChild(emailElement);
+        $location.path('/');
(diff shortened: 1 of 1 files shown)


Seidr:
I've updated this pull request to use AngularJS methods, instead of using vanilla JS. The method of deleting the current item may not be optimal - feedback always appreciated.

Grynn:
Nice work, thanks Seidr. 

Now instead of selecting the the 'parent' element, it would be awesome if we could select the previous item if there was one.

Seidr:
No problem! Thanks for the awesome project!

I suppose you could safely redirect to history-1, if it is not null, and redirect to '/' if it is null. What do you think?

djfarrelly:
This looks like an awesome change @seidr! I'm checking this out from my phone, but will pull it down and test it out tomorrow. Thanks!

Grynn:
heh! I'm just a lay 'fan' of the project too ;) 

Would it not be better to do something like:
$location.path('/email/' + $scope.items[idx-1].id) where idx is the index of current selection? (i.e. index of the item just deleted)?

I am an Angular newb so could be waaay off base etc.

djfarrelly:
Thanks for the work guys! Will get this out in the next release :+1:

Seidr:
Awesome, thanks! :)

PR #67, opened 2015-05-21
Add .listen() to the API example

gpbl:
As I understand it, calling `listen()` wasn't needed before v0.10... this change to the README may help to avoid confusion.


djfarrelly:
Awesome, thanks for this @gpbl!

Issue #70, opened 2015-06-16
Emit sockets event on delete


cbdabner:
Should _delete_ and _delete all_ emit events so that the frontend can react to deleted mail.

Currently, if one user deletes an item, other users still see it in the list and when they click on it they get shown the info landing page, and it's not intuitive that it's been deleted. They have to refresh the page to see the mail gone.

Maybe if a sockets event was emitted, the items could be removed from everyone's list.
Unless the user is actively looking at that item, in which case maybe you want some additional logic (the view that is currently shown can be left on screen until navigating elsewhere).

In fact, it looks like calling the Refresh logic is probably the desired result on delete.

Let me know your thoughts, and I might make a pull request later.

djfarrelly:
This is a good point. I guess I've never used MailDev w/ multiple users/clients that needed to be in sync. I would definitely be open to a pull request that emitted an event to update the front-end!

Just for my own curiosity, how are you using MailDev that multiple users are using the interface simultaneously? I love to hear how people end up using the project :smile:

cbdabner:
Previously, I worked in a private dev company where a hotfix was being developed for a client project, and the database contained some prod data. Emails were sent to public customers detailing scheduling of events, which wasn't pretty (not my work, although I'm sure I could have made the same mistake). So I setup MailCatcher at the time to help prevent that kind of thing.

Now, I'm working in the IT team (100 staff or so) for a larger public company who does some in-house development, but more often works on projects implementing various third-party software/solutions. There's a lot of processes etc that makes the potential for similar events less likely, but even so it's nice to avoid spamming your own inbox when projects are in dev/test.

I've setup MailDev in a central location, and have it starting up on a Windows Service (not allowed any Linux servers here). I haven't really spread the word much yet while I had a look at how it works. I haven't really looked into what occurs resource-wise etc.

The plan is to configure all projects in dev/test to use it, in conjunction with editing the dev/test servers' hosts file to remap any real SMTP to MailDev (in case someone misses the configuration). So in terms of multiple users, at any one time there could be up to maybe a dozen projects semi-actively sending mail. Additionally, I would expect the Test teams to get good usage out of it, being able to see the _exact_ intention of an application instead of finding some way to force an event to email themselves.

The Search box is probably good enough to differentiate between projects. It's pretty common for each project to use a custom _From_ address. Again, I haven't looked into the code, but it looks like when a new mail comes in, it adheres to the active search condition.

[...]

PR #71, opened 2015-06-17
Emit socket event on delete djfarrelly/MailDev#70

Seidr:
Essentially moved the behaviour of the delete mail to a socket event, instead of straight after the delete is submitted. This means an email will only be removed from the GUI if the server has actually deleted it, and also means the delete will be broadcast to other users.

### DIFF
--- a/app/scripts/controllers/main.js
+++ b/app/scripts/controllers/main.js
@@ -57,6 +57,25 @@ app.controller('MainCtrl', [
       }
 
     });
+    
+    $rootScope.$on('deleteMail', function(e, email) {
+      var idx = $scope.items.reduce(function(p, c, i){
+        if (p !== 0) return p;
+        return c.id === email.id ? i : 0;
+      }, 0);
+
+      var nextIdx = $scope.items.length === 1 ? null :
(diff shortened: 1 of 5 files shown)


Seidr:
As an after thought, I wonder events should be emitted when an email is read, otherwise a disparity could be introduced when one user reads an email. The unread count would then be out of sync between clients.

app/scripts/controllers/main.js
```diff
@@ -57,6 +57,30 @@ app.controller('MainCtrl', [
...
+                      idx === 0 ? idx + 1 : idx - 1;
+        if (nextIdx !== null) {
+          $location.path('/email/' + $scope.items[nextIdx].id);
+        } else {
+          $location.path('/');
+        }
+        
+        $scope.items.splice(idx, 1);
+        countUnread();
+        $scope.$apply();
```
djfarrelly: I like this :+1:

djfarrelly:
@Seidr you're awesome :smile:

Seidr:
Thanks :D I find MailDev is very fun to work with, and nicely written :)

Issue #30, opened 2014-10-05
Replace simplesmtp as SMTP server


djfarrelly:
simplesmtp is deprecated and will not be Node 0.12 compatible, ~~we need to switch to Haraka: https://github.com/baudehlo/Haraka~~

**Edit:** We could also switch to https://github.com/substack/node-smtp-protocol and https://github.com/andris9/smtp-connection, Haraka is a bit complicated for what the project needs

djfarrelly:
Found an example gist that could help: https://gist.github.com/steffenmllr/cb3d3af13d0263adf268

djfarrelly:
From basic testing...it seems that MailDev is still working running on the latest [io.js](https://github.com/iojs/io.js) 1.0.0 alpha. We may have more time!

djfarrelly:
Since Andris has now created [smtp-server](https://github.com/andris9/smtp-server) in addition to [smtp-connection](https://github.com/andris9/smtp-connection), I think this switch might be *way* easier to do! Combining those two to replace simplesmtp seems like the best plan here.

Seidr:
@djfarrelly I'm going to do some playing at lunch today, and see how close I can get it to the current feature set. *takes a deep breath*

PR #74, opened 2015-06-19
Replace simplesmtp with smtp-server and smtp-connection djfarrelly/MailDev#30

Seidr:
RE djfarrelly/MailDev#30

I've tested receiving emails with and without attachments, as well as relaying with and without authentication. All appears to work, however I'm not sure about the definitions of the callbacks in lib/mailserver.js around lines 417-466 (in particular mailConnectCallback and mailSendCallback). Open to suggestions as to how it can be improved - it just feels messy scoping them like that.

The change to smtp-server and smtp-connection does require NodeJS 0.12, as it depends upon Sets, which became available with 0.12.

### DIFF
--- a/lib/mailserver.js
+++ b/lib/mailserver.js
@@ -218,7 +218,7 @@ mailServer.listen = function(callback) {
  */
 
 mailServer.end = function(done){
-  smtp.end(done);
+  smtp.close(done);
 };
 
 
--- a/lib/mailserver.js
+++ b/lib/mailserver.js
@@ -3,7 +3,8 @@
  * MailDev - mailserver.js
  */
 
-var simplesmtp    = require('simplesmtp');
+var SMTPServer    = require('smtp-server').SMTPServer;
+var SMTPConnection = require('smtp-connection');
 var MailParser    = require('mailparser').MailParser;
 var events        = require('events');
 var logger        = require('./logger');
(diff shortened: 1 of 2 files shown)


Seidr:
Also, not sure about the level of verbosity at the moment. This could likely be reduced, and messages piped to logger.

Seidr:
Ah, of course, I'll take a look at the tests during my break. *facepalm*

Seidr:
Whoops - fixed the broken tests. The wrong method was being called in the end API.

Seidr:
Ok, I've fixed the broken 0.12 test, however the 0.10 test is failing due to not having the Set class available to it (as I said, introduced in 0.12). Not sure how to get this green again. Is that a project setting in github that you need to change?

Seidr:
I tried working around this by using the simplesets module, but that just resulting in further (and scarier) errors being thrown. Do we really need to bend over backwards to continue support of 0.10? Not sure of the market share vs 0.12.

djfarrelly:
Hey @Seidr, this was awesome to wake up to! I have a day of work ahead of me, but really want to dig into these changes this weekend, this looks great from a quick look.

It's my gut that market share of 0.10.x is still larger than 0.12.x, but potentially, MailDev could stop supporting Node 0.10.x with MailDev 0.11.0 and for Node 0.10 support a user can continue to use MailDev 0.10.x, I could maintain a branch for any breaking issue and publish patch releases if needed. With the impending Node/io.js merger, I would definitely like to test this on the latest io.js as I think that might hold more weight than Node 0.12 long term (just a personal opinion).

djfarrelly:
hey @Seidr! Man, I'm so sorry that this has sat on the shelf for all this time. The day job and vacation have taken priority recently. One fresh thought I had on this is to be backwards compatible requiring both packages and forking depending on which version of node the user is using. It would be more code and might be messy, but it would ensure it works for everyone no matter what version of Node they're developing. Then the 0.10.x code can be ripped out in the future when it's used less. I'm not really sure how other Node projects handle an issue like this, but I'm open to ideas!

I'm looking through your code right now and going to check more into how this could work.

djfarrelly:
Hey @Seidr, I did the fork I mentioned and added PR #77, what do you think? :)

Seidr:
Good idea re legacy support. Had a look at the PR and it's about as clean as it's going to get while retaining support for node 0.10.x.

Also, don't worry about delays - I know full well how 'real life' can take over from time to time :)

djfarrelly:
Yeah, I had thought about a further abstraction out of mailserver.js, but I think this is ok for now. It's not the cleanest code, but I'm thinking most consumers of the projects won't be digging into the codebase too often :)

djfarrelly:
Thanks again @Seidr! Merged this in, planning to release 0.11.0 today :+1:

Issue #80, opened 2015-07-25
Auto-relay mode


Ndrou:
Hi !

Idea:

Add the ability to configure maildev in auto relay mode. 
We could see in this case the emails that pass, but they would be transmitted to recipients.

Have a good day !

brunowego:
+1

Ndrou:
I will add to the pull request the ability to relay only certains emails with a system of include, exclude.

In the command line the ability to pass an option :
-auto-relay-include file.json -auto-relay-exclude file.json

The files will be as :

```javascript
[
  '*@gmail.com',
  'foo@bar.com'
]
```

PR #81, opened 2015-07-25
Add auto relay mode

Ndrou:
Adding an auto relay mode
See #80 for more details.


Ndrou:
Ok i added filter rules in addition to the auto relay mode.

djfarrelly:
Sorry for the delay here, this should be on npm now @Ndrou, thanks for the PR :+1:

Ndrou:
ok thanks :) have a good day ;)

PR #85, opened 2015-08-31
Adding async queue for relaying email in order to fix bug when bulk send

Ndrou:
When I sent 2 emails at the same time in MailDev, i had an SMTP error. 
So I delayed email sending through a queue.


djfarrelly:
Sorry for the delay on this one @Ndrou, thanks so much for your contribution, I'm getting this out in a 0.12.1 release ASAP!

Issue #86, opened 2015-09-04
postfix denies relayed mails


diresi:
I'm using the docker image of v0.12.0.

I'm inserting simple mails using the unix `mail` command:
``` bash
echo "teest" | mail -S 'smtp=smtp://localhost:1025' -s 'a subject' recipient@example.com
```

My outgoing configuration seems OK, but when I try to relay the above email postfix responds with
```
NOQUEUE: reject: RCPT from unknown[192.168.88.105]: 550 5.1.1 <null>: Recipient address rejected: User unknown in local recipient table; from=<> to=<null> proto=ESMTP helo=<[127.0.0.1]>
```

This appearently means that both `From:` and `To:` are empty.
I "fixed" this using a small patch, but I have no clue whether this is the correct approach ...

The patch is:
``` patch
diff --git a/lib/mailserver.js b/lib/mailserver.js
index 02d7211..b617e5a 100644
--- a/lib/mailserver.js
+++ b/lib/mailserver.js
@@ -255,7 +255,16 @@ function relayMail(idOrMailObject, autoRelay, done) {
 
       var mailSendCallback = function(err, info) {
 
-        var recipients = mail.envelope.to.map(function(toItem) { return toItem.address; });
+        var get_address = function(address) {
+          email = address.address;
+          if (email === undefined) {
+            email = address;
+          }
+          return email;
+        }
+
+        var recipients = mail.envelope.to.map(get_address);
+        var sender = get_address(mail.envelope.from);
 
         if (autoRelay && mailServer.autoRelayRules) {
           recipients = getAutoRelayableRecipients(recipients);
@@ -267,7 +276,7 @@ function relayMail(idOrMailObject, autoRelay, done) {
         }
 
         mailServer.client.send({
-          from: mail.envelope.from.address,
+          from: sender,
           to: recipients
         }, rawEmailStream, function (err, info) {
           if (err) {
```

I'd be happy to create a merge request if you're willing to accept this patch.

diresi:
Hmm. Although relaying works (the mail is delivered) MailDev exits several minutes later. The full output is:
```
Temporary directory created at /tmp/maildev
Temporary directory created at /tmp/maildev/1
MailDev outgoing SMTP Server ${SMTP_SERVER}:25 (user:undefined, pass:undefined, secure:no)
MailDev app running at 0.0.0.0:80
MailDev SMTP Server running at 0.0.0.0:25



events.js:72
        throw er; // Unhandled 'error' event
              ^
[...]

PR #87, opened 2015-10-06
Connection close and relay

donut87:
Hi,

these two commits allow us to relay incoming mails. diresis commit ensures, that there is no "null" recipient or sender and the other commit closes the smtp connection after sending an email. This was necessary because we always got an error when the connection was closed by the smtp server.

Cheers

Christian

### DIFF
--- a/lib/mailserver.js
+++ b/lib/mailserver.js
@@ -255,7 +255,16 @@ function relayMail(idOrMailObject, autoRelay, done) {
 
       var mailSendCallback = function(err, info) {
 
-        var recipients = mail.envelope.to.map(function(toItem) { return toItem.address; });
+        var get_address = function(address) {
+          email = address.address;
+          if (email === undefined) {
+            email = address;
+          }
+          return email;
+        }
+
+        var recipients = mail.envelope.to.map(get_address);
+        var sender = get_address(mail.envelope.from);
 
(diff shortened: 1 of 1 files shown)


diresi:
Works for me, thanks a lot!

djfarrelly:
So sorry about the delay on this one @donut87 and @diresi! I really appreciate this awesome fix and will be pushing out 0.12.1 to npm ASAP :+1:

djfarrelly:
Hi @donut87! I've pulled this down and have done a bunch of testing and when using Gmail, this will work on the first relay, but will fail on any any future relays. It seems the reconnection needed to be changed to `config.pass && config.user`.

I'm curious - why did you choose to instantiate a new SMTPConnection object in `resetRelayConfiguration` instead of letting it quit then reconnect with client.connect? I'm running into more issues when combined with the changes in #85 - Thanks for your help!

diresi:
@djfarrelly I've started this in #86, maybe the description there helps?

djfarrelly:
Ah ok, thanks @diresi - With the timeout issue I'm thinking a better flow for maildev might be `create client > connect > login > send > quit` vs. `connect > login > send > quit > create client`. Long term there could be a way to test if the connection was still active vs. quitting + re-connecting :) Thanks for your help!

Issue #100, opened 2016-01-05
Options on Docker image


Soullivaneuh:
Your application allow to set many options.

Would be great to be able to add it on the docker image, like the verbose one for example.

djfarrelly:
Hi @Soullivaneuh! You should be able to override the command when running the container, i.e.

```
$ docker run -P 80:80 -P 25:25 djfarrelly/maildev bin/maildev --web 80 --smtp 25 --verbose
```

Soullivaneuh:
Well, we ca indeed do that. But it's not really pretty IMO because you have to override all the command line with default options.

If on another version, you change the default command line for good reason, I would not see it.

Soullivaneuh:
FYI, I'm running maildev with docker-compose:

```yml
maildev:
  image: djfarrelly/maildev
  ports:
    - '1080:80'
```

Would be great to deal with env section:

```yml
maildev:
  image: djfarrelly/maildev
  ports:
    - '1080:42'
  environment:
    - HTTP_PORT=25
    - SMTP_PORT=yes
    - VERBOSE=yes
```

With this method, env variable will be available for linked container too, this can provide useful information.

Are you interested of a PoC PR?

djfarrelly:
Hey @Soullivaneuh, thanks for your thoughts and effort on this one. I'm a bit hesitant on this one as it simply adds another configuration mechanism where I don't personally see that it's too bad here to use the arguments as one would configure MailDev on the command line.

```yml
maildev:
  image: djfarrelly/maildev
  cmd: bin/maildev --web 80 --smtp 25 --verbose
  ports:
    - '1080:80'
```
or alternatively...
```yml
  cmd: bin/maildev
    --web 80 
    --smtp 25 
    --verbose
```

The approach may be "less pretty," but it's easy to maintain consistency with the project. I'll keep this open in case others want to chime in and share thoughts

Soullivaneuh:
@djfarrelly Yes, this is what I used. But having options permit more flexibility.

If, for reason, one of my developers has to change only the web port, he actually has to override the full command and will get rid of any update about it.

With options, it will be able to change only the port and don't touch to the raw command.

So I really think having options give some advantages. Furthermore, most of the official Docker images use this way,

Theon01:
Just a heads up :
`command` should be used instead of `cmd` in the most recent version of docker compose.

curry684:
[...]

Issue #93, opened 2015-11-07
Add maildev.json for quick project configuration


djfarrelly:
So people can share maildev configurations across projects w/out the need for command line options. This would also deprecate the `rules.json` that a maildev user can use. Idea:

**maildev.json**
```
{
  "smtp": 1025,
  "web": 1080,
  "autoRelayRules": [
    { "allow": "*@allsafe.com" },
    { "deny":  "elliot@allsafe.com" },
    { "allow": "*@evilcorp.com" },
  ]
}
```

Thoughts or input are greatly appreciated!

Soullivaneuh:
:+1: 

This could be very useful because we can link our maildev.json configuration file with docker-compose (related to #100).

Maybe a `.maildev.json` would be better to not have the root project full of config files.

Another idea: Add an option to the binary to specify the json file to use.

Soullivaneuh:
Which npm package could we use to parse a configuration file? Maybe I can work on it.

Soullivaneuh:
Maybe this one: https://github.com/pilwon/node-get-config

Soullivaneuh:
Actually, how the `rules.json` system works?

djfarrelly:
Implemented in 3.0. MailDev now has a full config-file subsystem in `packages/cli/src/config/`:

- Auto-discovered files (walking up from cwd): `.maildevrc.json`, `maildev.config.ts`, `maildev.config.js`, `maildev.config.mjs`, `maildev.config.cjs`
- Explicit path via `--config <path>`
- `maildev init` scaffolds a starter config (`--json` for `.maildevrc.json`)
- Precedence: CLI args > env vars > config file > defaults

Auto-relay allow/deny rules are also configurable (via `--auto-relay-rules` and config). The filename ended up as `.maildevrc.json` rather than `maildev.json`, but the shareable-project-config idea from this issue is fully delivered. Closing as completed.

PR #103, opened 2016-01-19
Docker verbose option

Soullivaneuh:
Here is a PoC as suggested on https://github.com/djfarrelly/MailDev/issues/100#issuecomment-172815644.

I just added verbose options but we can add more.

But I agree that having a `maildev.json` configuration file (https://github.com/djfarrelly/MailDev/issues/93) would be easier. With that, we will just have to link our config file to the docker container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,11 @@
 FROM node:5-onbuild
+
+ENV VERBOSE 0
+
 EXPOSE 80 25
-CMD ["bin/maildev", "--web", "80", "--smtp", "25"]
+
+ADD entrypoint.sh /
+RUN chmod u+x /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
+
+CMD ["sh", "-c", "bin/maildev --web 80 --smtp 25"]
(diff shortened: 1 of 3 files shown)


Soullivaneuh:
Another possibility: Let user set a `OPTS` env var that will be added to the default (web and smtp options should not be changed on docker due to exposed ports).

What do you think?

djfarrelly:
For this PR, I do think my feelings are still the same as [this comment](https://github.com/djfarrelly/MailDev/issues/100#issuecomment-184511038), but your point about using an `OPTS` or similar environment variable for additional args could be a cool point to meet halfway! I wonder, could we use an environment variable in the `CMD` itself and skip the `entrypoint.sh`? 

```Dockerfile
ENV OPTS=""
CMD ["sh", "-c", "bin/maildev --web 80 --smtp 25", "$OPTS"]
```

Soullivaneuh:
Hi @djfarrelly, sorry for the delay.

I think your method is less reliable about option description, but we could do that.

Do you want me to update this PR accordingly to your demand?

Issue #106, opened 2016-02-27
Getting "Cannot GET /" when trying to access server


atrauzzi:
I've got maildev going in a setup right now where any time I try to access the interface, I get the message "Cannot GET /" in my browser.

There's no console output, not sure what might be happening in this situation.  Any hints on what it might be based on this message?

dataich:
+1

m-nagae:
:+1:

akivajgordon:
+1

liul85:
+1

liul85:
use node version of v0.10.33 fix this issue for me on mac.
previous version is 5.7.0 installed by homebrew.

robinparisi:
Same here with node 5.7.0

vangale:
Node 5.6.0 worked fine,  but I have the same problem with 5.7.0.  Unfortunately I can't downgrade versions from the nodesource repo (they've already deleted 5.6.0 from debian repo) so I'll need to find another source or build it.

Joshaharonoff:
+1

ibmz:
+1

TechFounder:
+1 !!!

cyrosy:
Same error with node 5.7.1

stianpr:
+1

djfarrelly:
Thanks to all for all the +1's. Bug fix release Maildev `0.12.2` has been published on npm w/ @jameshhx's wonderful fix!

fdanielsen:
Maildev `0.12.2` indeed works, but the same error is back in Maildev `0.13.0` at least when running with node 5.9.0.

djfarrelly:
Hi @fdanielsen, I've just installed node 5.9.1 with nvm then installed maildev@0.13 and everything seems to work alright - Can you double check your versions or possibly try Node 5.9.1? Thanks!

jameshhx:
I just tried with version `5.9.0` of node, on version `0.13.0` of Maildev, and it looks to be working fine. Are you sure the most recent version of Maildev was installed under the environment for version `5.9.0` of node?

```
21:28 ~: nvm install v5.9.0
######################################################################## 100.0%
Checksums empty
Now using node v5.9.0 (npm v3.7.3)
21:28 ~: npm install -g maildev
/Users/james.hooker/.nvm/versions/node/v5.9.0/bin/maildev -> /Users/james.hooker/.nvm/versions/node/v5.9.0/lib/node_modules/maildev/bin/maildev
[...]

PR #107, opened 2016-03-07
Fixing path to static directory

jameshhx:
This PR addresses the issue seen in more modern node versions. The path for the static files appears to be traversing one level too many. I've removed the extra level of traversal and am making use of the `path` module to join the paths together.

I've tested this solution on versions v0.10.40, v0.12.7 and v5.7.1. In all three versions the GUI displays as expected. This was tested by updating to the target `node` versions and installing the `maildev` package on each using `npm`. The change was also tested by cloning the project down in each version, and then running `bin/maildev`.

This resolves issue #106 

p.s. Nice to see this is still going @djfarrelly - love the additions! :+1:

### DIFF
--- a/lib/web.js
+++ b/lib/web.js
@@ -10,7 +10,7 @@ var io = require('socket.io').listen(server);
 var routes = require('./routes');
 var auth = require('./auth');
 var logger = require('./logger');
-
+var path = require('path');
 
 /**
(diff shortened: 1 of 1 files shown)


djfarrelly:
Thanks so much @jameshhx, you're the man!!

PR #111, opened 2016-03-22
Fix attachments when using nodemailer

ebourmalo:
Small patch to fix the error occurring when we add attachments to an email.

When maildev tries to [save the mail](https://github.com/djfarrelly/MailDev/blob/master/lib/mailserver.js#L62) by copying the mail object (returned by MailParser), we got an error `Cannot convert circular JSON`. 
It's due to the stream object left in attachments of the mail object.

I don't really know if we should blame MailParser for a breaking-change minor update, but this stream object should not be given, especially if we explicitly process the attachments with a dedicated stream (as we do with maildev).

Usage context:
- node 5.1 / es6
- [nodemailer](https://github.com/nodemailer/nodemailer)
- transporter: [amazon SES](https://github.com/andris9/nodemailer-ses-transport)
- attachments: base64 string

This patch fix the case but not sure if it can be a long-term fix, what do you think @djfarrelly ?

### DIFF
--- a/lib/mailserver.js
+++ b/lib/mailserver.js
@@ -59,6 +59,14 @@ function makeId(){
 
 // Save an email object on stream end
 function saveEmail(id, envelope, mailObject){
+  
+  // remove stream object from attachments (fix the JSON.stringify)
+  if (mailObject.attachments instanceof Array) {
+    mailObject.attachments.forEach((attachment) => {
+      delete attachment.stream;
+    });
+  }
+
   var object = clone(mailObject);
 
   object.id = id;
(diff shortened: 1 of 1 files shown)


ebourmalo:
@djfarrelly Do you have any plan for this PR or looking for another way to fix the bug?

djfarrelly:
Hey @ebourmalo! I've been traveling a bit and sorry I didn't get to this PR yet. This feels like a good fix to me, I'll merge this and release `0.13.1` when I get back to solid wifi today!

djfarrelly:
This has been rebased into master (see https://github.com/djfarrelly/MailDev/commit/55476caad998fbddd90effbfdb379d41ab201bfd) and will be out in the 0.13.1 release 😄

ebourmalo:
Awesome, thank you @djfarrelly :)

PR #112, opened 2016-03-28
Reduce the size of the Docker image by switching to alpine base image

tcheneau:
Switch from node:5-slim to alpine:3.3 to reduce the image size (~270MB to ~42MB).

I also apply some cleanup commands so as to decrease the overall size of the image.

To be fair, I started this work because the standard image was segfaulting on my system (I wonder if it does that to other) and I didn't wanted to mess with the node image.


djfarrelly:
Thanks for this work @tcheneau!! This is great, just tested locally and will push to Dockerhub with 0.13.1 😄

PR #117, opened 2016-05-06
fix all inline images

### LINKED ISSUE #116: inline images are only loaded once
maildev replaces "src" attributes for inline images, but the regexp misses the global flag, thus fixing only the first occurrence of an inlined image.

diresi:
using the global flag on the search regexp applies replacement for all
matches in the document.

fixes #116


djfarrelly:
Wow, what a silly mistake by me 😄  Thanks for the fix @diresi! 👍

PR #131, opened 2016-07-26
Fix relative path to relay endpoint

oowl:
It was becoming wrong url when `basePathname` param was used.


djfarrelly:
Merged in #133

Issue #127, opened 2016-07-06
Add an option to keep only the last n messages


arnou:
To avoid memory problems, limit the number of messages in the box ( awaiting permanent storage ) .

Example: " --limit-messages 1000"

djfarrelly:
I like this idea. I've been messing around a bit locally with this, hoping to have something soon!

kbudde:
Hi, 
any updates on this issue? 
It would be really nice to have this option.

BR

Kris

alvillain:
Hi all,

@djfarrelly Do you need something before merging current PR? It is opened since 31st October 2016.

alvillain:
Hi @djfarrelly,
Any updates there? Do you plan to include the feature to the next release?
BR, Alex.

oktapodia:
Thanks for your issue, we will look into it :)

djfarrelly:
Still wanted and not yet exposed to users, but there's already a hook to build on: the in-memory store supports a `maxEmails` ring buffer that evicts the oldest message once the cap is exceeded (`MemoryStorage`, `packages/core/src/storage/memory.ts`). It defaults to `0` (unlimited) and isn't wired to anything.

To close this out, the work is: add a CLI flag / config field / env var (e.g. `--max-messages <n>`) and thread it into storage construction in `packages/cli/src/server/orchestrator.ts`. Note `FileStorage` has no equivalent eviction yet, so persistent storage would need a matching implementation. Relabeling `enhancement`; PR welcome.

djfarrelly:
Cross-referencing #258, which asks for the same area from another angle. Two retention modes worth covering when this is implemented:

1. **Count-based (this issue):** keep only the last N messages (FIFO-evict oldest) — the `maxEmails` ring buffer in `MemoryStorage` already does this; it just needs a CLI/config/env flag wired through `orchestrator.ts` (and an equivalent for `FileStorage`).
2. **Age-based (from #258):** delete messages older than a given period (e.g. `--max-age 14d`).

PR #140, opened 2016-10-31
#127 

p610:
Storage moved to separate js file - mainly because it's easier to test. Instead of using a normal js array a custom implementation of push method is provided. A storage itself uses an event emiter to communicate with the rest of the system.

It's not an ideal solution but provides the functionality which is needed. The storage might be use in the future to provide a proper separation of mailserver and storage functionalities. This should provide an abstraction layer between multiple storages and mailserver. Anyway it's a bigger refactoring effort.

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,6 +34,7 @@ module.exports = function(config) {
       .option('--auto-relay-rules <file>', 'Filter rules for auto relay mode')
       .option('--incoming-user <user>', 'SMTP user for incoming emails')
       .option('--incoming-pass <pass>', 'SMTP password for incoming emails')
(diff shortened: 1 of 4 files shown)


joh1234:
Update README
https://goo.gl/JBljdR

or 
http://bit.ly/2dK6Ojk

oktapodia:
Thanks for your work, can you please update your branch?

oktapodia:
Fix #127

Polosa (approved):

oltionzefi (approved):

ephrin (approved):

PR #143, opened 2016-10-31
Feature/swagger support

p610:
Added support for swagger. Swagger document is manually managed. Does not contain body definitions at this stage. Might be worth to add them later on as it can be used for sdk generation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,6 @@ node_js:
   - "0.10"
 after_success:
   - bash <(curl -s https://codecov.io/bash)
+script:
+  - npm run validate:swagger
+  - npm test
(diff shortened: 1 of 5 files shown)


donut87:
Hey. I just had a brief look over what swagger does. So here is my question: What problem does swagger solve in this project?
To explain: I am not against using frameworks and libraries where they solve problems or satisfy needs. Adding a framework just because we can might be counterproductive.

djfarrelly:
I agree w/ @donut87. Going to close this for now. Although I appreciate what it does and your work @p610, I don't feel like there is enough case for Swagger right now in MailDev. Thank you for your contribution!!

PR #144, opened 2016-10-31
Feature/swagger doc gen

p610:
Additional script that generates markdown documentation from swagger.yaml file. Might replace api.md in the feature. It should be always up to data as it's a generated documentation. 

It's nice to have in my opinion but not necessary. 

btw. this PR heavily relays on previous one - swagger integration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,5 @@ node_js:
 after_success:
   - bash <(curl -s https://codecov.io/bash)
 script:
-  - npm run validate:swagger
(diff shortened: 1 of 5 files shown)


joh1234:
Update README
https://goo.gl/JBljdR

or 
http://bit.ly/2dK6Ojk

djfarrelly:
See: https://github.com/djfarrelly/MailDev/pull/143#issuecomment-299671527

PR #146, opened 2016-11-09
Change source request to be lazy

cbdabner:
The server side uses MailParser's `streamAttachment` option to improve performance on emails with large attachments. However, the client app requests `/email/:id/html` and `/email:id/source` on item load. This moves the source call to only occur when Display --> Source is clicked.

### DIFF
--- a/app/scripts/controllers/item.js
+++ b/app/scripts/controllers/item.js
@@ -16,7 +16,6 @@ app.controller('ItemCtrl', [
       Email.get({ id: $routeParams.itemId }, function(email) {
 
         $scope.item = new Email(email);
-        $scope.rawEmail = 'email/' + $scope.item.id + '/source';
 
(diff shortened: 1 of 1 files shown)


joh1234:
Update README
https://goo.gl/JBljdR

or 
http://bit.ly/2dK6Ojk

cbdabner:
My comment probably wasn't clear enough. The performance cost was around the loading of /source.

/Source can include large attachments within the .eml, and can cause the UI to become unresponsive while it loads. This occurs everytime when viewing an email. For us it was a 2-5sec hang when navigating between emails with larger attachments.

Considering it's quite rare to actually navigate to the /source view and the change to make it load only when viewed was very minor, it seemed like a reasonable quick fix. I'm not sure if you should be able to make use of the stream attachment feature or something more optimal.

PR #147, opened 2016-12-03
Add --hide-extensions option that can be used to not advertise certain SMTP extensions

Deoxyribonucleic:
I needed to test ASCII-only headers so I added this option to be able to keep the server from sending the  `SMTPUTF8` extension in the `EHLO` response.

Functionality for hiding this extension (as well as `PIPELINING` and `8BITMIME`) was added in `smtp-server` 1.10.0 so while I was at it I upgraded the `smtp-server` dependency from 1.4.0 to 1.16.1. All tests still pass after adding a newline to the end of the test mail body (see [`smtp-server` 1.11.2](https://github.com/andris9/smtp-server/blob/master/CHANGELOG.md#v1112-2016-07-15)).

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,25 +26,29 @@ For convenient use with Grunt, try [grunt-maildev](https://github.com/xavierprio
 
     maildev [options]
 
-      -h, --help                    output usage information
-      -V, --version                 output the version number
-      -s, --smtp <port>             SMTP port to catch emails [1025]
-      -w, --web <port>              Port to run the Web GUI [1080]
-      --ip <ip address>             IP Address to bind services to [0.0.0.0]
-      --outgoing-host <host>        SMTP host for outgoing emails
-      --outgoing-port <port>        SMTP port for outgoing emails
-      --outgoing-user <user>        SMTP user for outgoing emails
-      --outgoing-pass <pass>        SMTP password for outgoing emails
-      --outgoing-secure             Use SMTP SSL for outgoing emails
-      --auto-relay                  Use auto relay mode
-      --auto-relay-rules <file>     Filter rules for auto relay mode
-      --incoming-user <user>        SMTP user for incoming emails
-      --incoming-pass <pass>        SMTP password for incoming emails
-      --web-ip <ip address>         IP Address to bind HTTP service to, defaults to --ip
-      --web-user <user>             HTTP basic auth username
-      --web-pass <pass>             HTTP basic auth password
-      -o, --open                    Open the Web GUI after startup
-      -v, --verbose
+      -h, --help                      output usage information
+      -V, --version                   output the version number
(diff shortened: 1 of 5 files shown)


Deoxyribonucleic:
Well, they ran on my machine at least...

jonaz:
@djfarrelly status on this one? I would really need this aswell to turn off STARTTLS so i dont have to configure all my clients to ignoreTlsSecurity.

README.md
```diff
@@ -26,25 +26,29 @@ For convenient use with Grunt, try [grunt-maildev](https://github.com/xavierprio
...
+      --outgoing-host <host>          SMTP host for outgoing emails
+      --outgoing-port <port>          SMTP port for outgoing emails
+      --outgoing-user <user>          SMTP user for outgoing emails
+      --outgoing-pass <password>      SMTP password for outgoing emails
+      --outgoing-secure               Use SMTP SSL for outgoing emails
+      --auto-relay                    Use auto-relay mode
+      --auto-relay-rules <file>       Filter rules for auto relay mode
+      --incoming-user <user>          SMTP user for incoming emails
+      --incoming-pass <pass>          SMTP password for incoming emails
+      --web-ip <ip address>           IP Address to bind HTTP service to, defaults to --ip
```
donut87: Why did this "change"? I could not see whitespaces or anything...

djfarrelly: I think this is the indentation perhaps with the longer `--hide-extensions <extensions>` pushing things to the right?

lib/mailserver.js
```diff
@@ -258,14 +259,18 @@ mailServer.create = function(port, host, user, password) {
...
       incomingUser: user,
       incomingPassword: password,
       onAuth: authorizeUser,
       onData: handleDataStream,
       logger: false,
       disabledCommands: !!(user && password)?['STARTTLS']:['AUTH']
-    });
+    };
+
+    addHideExtensionOptions(smtpServerConfig, hideExtensions);
```
djfarrelly: This is a bit of a style thing, but I'd prefer a bit more functional way of writing this, perhaps something like:

```js
const smtpExtensionConfig = getValidExtensionOptions(hideExtensions)
smtpServerConfig = Object.assign({}, smtpServerConfig, smtpExtensionConfig)
```
That makes the code more testable. If we mutate the object passed by reference it isn't clear that it's changing the object. Can we refactor this?

djfarrelly: I made these changes here: https://github.com/djfarrelly/MailDev/pull/169/commits/472d71312e4e85b33aed527c89cbaee2a8a01003

djfarrelly:
@donut87, I do agree, the options are getting a bit out of hand, perhaps instead of arguments we can use a configuration object? This would be good think about and improve!! 😄

Issue #156, opened 2017-02-01
Does not process SIGINT properly


jsonxr:
When running via docker command line, it does not shut down when I press "CTRL-C".  

    docker run --name=mail --rm -p "1080:80" -p "25:25" djfarrelly/maildev

I have to type:

    docker rm -f mail

To force it to shut down.

Adding 

    process.on('SIGINT', process.exit); // Handle ctrl-c when running via docker run
    process.on('SIGTERM', process.exit); // Handle when we docker rm the container

To bin/maildev would handle this

jsonxr:
Pull request to fix:
https://github.com/djfarrelly/MailDev/pull/157

djfarrelly:
I have a feeling that this is happening because the connections are staying open, possibly with Socket.io. I'm looking into this.

PR #157, opened 2017-02-01
Issue 156: Shut down when receiving ctrl-c

jsonxr:
This provides a fix for Issue 156 where you can't shut it down when running with docker.

### DIFF
--- a/bin/maildev
+++ b/bin/maildev
@@ -3,6 +3,9 @@
 var path = require('path');
 var fs = require('fs');
 
+process.on('SIGINT', process.exit); // Handle ctrl-c when running via docker run
+process.on('SIGTERM', process.exit); // Handle when we docker rm the container
+
 var root = path.join(path.dirname(fs.realpathSync(__filename)), '../');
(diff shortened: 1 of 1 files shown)


bin/maildev
```diff
@@ -3,6 +3,9 @@
 var path = require('path');
 var fs = require('fs');
 
+process.on('SIGINT', process.exit); // Handle ctrl-c when running via docker run
+process.on('SIGTERM', process.exit); // Handle when we docker rm the container
```
donut87: Please don't do that here. Or maybe, do that here, but please have a look at the index.js. There is already a listener for SIGTERM.

djfarrelly:
I've been testing locally on this and I still see the issue from #156 where it doesn't shutdown because the webserver is hanging on it's closing. This [callback](https://github.com/djfarrelly/MailDev/blob/77392aceecd8d4edf2ae37c997909bade073b3ca/lib/web.js#L87) never gets called.

djfarrelly:
Addressed in #167. Thanks for contributing @jsonxr!!

Issue #159, opened 2017-02-23
Crashes when smtp port set to anything below 1024


realjax:
MailDev app running at 0.0.0.0:1080
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: listen EACCES 0.0.0.0:1023
    at Object.exports._errnoException (util.js:1022:11)
    at exports._exceptionWithHostPort (util.js:1045:20)
    at Server._listen2 (net.js:1246:19)
    at listen (net.js:1295:10)
    at net.js:1405:9
    at _combinedTickCallback (internal/process/next_tick.js:77:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
    at Module.runMain (module.js:606:11)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149:9)

donut87:
Just from the top of my head...
Do you have the permission to use this port? Try running maildev as root?

realjax:
Haven't tried as far as I can  remember. But it shouldn't crash nonetheless.

Christian Dohnert <notifications@github.com> schreef op 18 maart 2017 17:02:10 CET:
>Just from the top of my head...
>Do you have the permission to use this port? Try running maildev as
>root?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/djfarrelly/MailDev/issues/159#issuecomment-287555396

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

donut87:
I've just confirmed your problem and my theory. Yes if you run this as root, port 1023 will work just fine. Except if the port is taken. I also think you are right. This should end gracefully and not shut down with a strange looking error message, but with some useful information (could not bind port X). Maybe I'll be able to do fix this in the next days.

realjax:
Cool, thanks!

Christian Dohnert <notifications@github.com> schreef op 18 maart 2017 23:16:33 CET:
>I've just confirmed your problem and my theory. Yes if you run this as
>root, port 1023 will work just fine. Except if the port is taken. I
>also think you are right. This should end gracefully and not shut down
>with a strange looking error message, but with some useful information
>(could not bind port X). Maybe I'll be able to do fix this in the next
>days.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/djfarrelly/MailDev/issues/159#issuecomment-287578834

-- 
[...]

PR #162, opened 2017-03-18
Trying to solve issue #159

donut87:
We now shout down the server with an error message that it could not be started, if MailDev was unable to bind the ports. See #159

### DIFF
--- a/lib/mailserver.js
+++ b/lib/mailserver.js
@@ -300,7 +300,7 @@ mailServer.listen = function(callback) {
   });
 
   smtp.on("error", function(e) {
-    throw new Error("Could not bind address " + e.address + ":" + e.port + "\nPort already in use or insufficient rights to bind port");
+    throw new Error("Could not start mail server on " + e.address + ":" + e.port + "\nPort already in use or insufficient rights to bind port");
   });
 };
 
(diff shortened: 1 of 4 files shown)


donut87:
It seems, that I will have to write tests for this code. Not now. I am too tired and have zero knowledge on writing tests in Nodejs.

donut87:
Works on my machine...
Now I am back to guess work and this might generate some mails...

donut87:
Ok. The last error was not that hard. There was no server running so no call to "mailServer.end" was necessary. But again... worked on my machine...

lib/mailserver.js
```diff
@@ -25,7 +25,6 @@ var config = {};
 var store = [];
 var tempDir = path.join(os.tmpdir(), 'maildev', process.pid.toString());
 var eventEmitter = new events.EventEmitter();
-var smtp;
```
donut87: This is now a local variable, which I unfortunately forgot to define. Will do so.

package.json
```diff
@@ -38,7 +38,8 @@
     "smtp-connection": "2.3.1",
     "smtp-server": "1.4.0",
     "socket.io": "1.4.5",
-    "wildstring": "1.0.8"
+    "wildstring": "1.0.8",
+    "expect": "1.20.2"
```
donut87: Aggreed. Will do so.

Issue #75, opened 2015-06-23
MailDev preventing Node from exiting?


mramato:
Hopefully there is a simple answer to my problem, but I couldn't find anything from searching past issues or the web.

I have a very simple Jasmine unit test set up where I programatically launch a MailDev server, send a few emails, validate they were received by MailDev and then shutdown.  However, the moment I introduce a `MailDev` server into my test, it prevents Node from exiting after the tests are complete.  Here's a reduced use case that causes the problem.  The "Do nothing" test passes, then the `Shut down` message gets printed, but it looks like MailDev still has active handles/requests open, preventing Node from exiting.

``` javascript
beforeAll(function (done) {
    mailServer = new MailDev({
        smtp: port,
        web: 8082,
        verbose: false
    });
    mailServer.listen(function (err) {
        if (err) {
            console.log(err);
            return done.fail();
        }
        console.log('Started up');
        done();
    });
});

afterAll(function (done) {
    mailServer.end(function () {
        console.log('Shut down');
        done();
    });
});

it("Do nothing", function (done) {
    done();
});
```

The moment I comment out the `listen` and `end` calls (still need to call `done`), the problem goes away, so I'm pretty sure the issue in in MailDev itself.  I hope I'm just doing something wrong on my end, but any help would be appreciated.  Thanks.

sponte:
I see the same behaviour. Have you ever got it working @mramato?

sponte:
After a quick look, the node process is not exiting because the web interface is still running after closing maildev server. I have added an option to disable it through configuration (```noWeb```). There is a pull request https://github.com/djfarrelly/MailDev/pull/165 waiting for approval but in the meantime I am using my fork directly.

```
npm install https://github.com/eu-evops/MailDev --save-dev
```

mramato:
> I see the same behaviour. Have you ever got it working @mramato?

Unfortunately not, I gave up and just mocked the actual sending of mail in my tests. If your PR gets merged, I would definitely give it a shot and revisit the issue.  Thanks for looking more into it!

sponte:
No worries. Just so you know it is working fine for me now with adding noWeb configuration option when instantiating new MailDev({ noWeb: true })

PR #165, opened 2017-04-24
Add an option to disable web interface

sponte:


sponte:
Fixes #75

index.js
```diff
@@ -72,20 +73,24 @@ module.exports = function(config) {
...
+    }
   }
 
   process.on('SIGTERM', function () {
-    async.parallel([
-      mailserver.end,
-      web.close
-    ], function(err, results) {
+    var tasks = [mailserver.end];
+    if(web) {
```
djfarrelly: I think this will always be truthy as it's the module's export object. It could be useful to have a method like `web.isListening()`?

Now that I think of it, [the `createServer` call in web.js](https://github.com/djfarrelly/MailDev/blob/77392aceecd8d4edf2ae37c997909bade073b3ca/lib/web.js#L8) should likely be moved into the `start` function then the `isListening()` function can check if `server` is defined or not. If you'd like to make this change, that works with me or I can hack it together after this is merged. Let me know what you think!

djfarrelly:
@sponte, I made this change I was talking about here: https://github.com/djfarrelly/MailDev/commit/0fe23cae52bc0eb639a1e6acea6b6c7d20d096d2

This will be in the next release! Thanks for contributing 😄

sponte:
@djfarrelly sorry for not replying, got lost in my inbox. Thanks for making adjustments and accepting pull request. All the best

PR #171, opened 2017-05-08
Upgrade mailparser and smtp-server [Requires Node 6]

djfarrelly:
### Purpose
To update core packages for upstream bug fixes. Includes various changes required to upgrade mailparser.

### Notes
I've decided to shelf this one as upgrading mailparser requires Node 6.x (spread operator) and I don't want to cut off Node 4.x right now. Comments welcome and appreciated.


djfarrelly:
This is very outdated and has since been upgraded.

PR #178, opened 2017-06-03
Passing of basepath changed to a require call

### LINKED ISSUE #176: basepath failing when looking for socket.io
Hey there,

I used the `--base-pathname` and all the links works except when searching for the socket.io files

screen shot 2017-05-25 at 11 53 29

I believe the issue and fix is described here, of adding the `basepath` information to the `require` call of **socket.io**

https://stackoverflow.com/questions/12824612/change-socket-io-static-file-serving-path

I'll try and submit a PR soon if I get time, but just letting y'all know

ObjectiveTruth:
Wasn't working because io.path doesn't work when attaching a server to an already defined socket.io object
closes #176

### DIFF
--- a/lib/web.js
+++ b/lib/web.js
@@ -6,7 +6,7 @@
 
 const express = require('express')
 const http = require('http')
-const io = require('socket.io')()
+let io = require('socket.io')()
 const routes = require('./routes')
 const auth = require('./auth')
 const logger = require('./logger')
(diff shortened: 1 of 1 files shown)


ObjectiveTruth:
@djfarrelly

ObjectiveTruth:
@djfarrelly pls pls

lib/web.js
```diff
@@ -80,9 +80,7 @@ web.start = function (port, host, mailserver, user, password, basePathname) {
   }
 
   if (basePathname) {
-    io.path(basePathname + '/socket.io')
-  } else {
-    basePathname = '/'
+    io = require('socket.io')({ path: basePathname + '/socket.io' })
   }
 
   app.use(basePathname, express.static(path.join(__dirname, '../app')))
```
djfarrelly: Without the `basePathname = '/'` line the app will not start up without the basePathname in this line: `app.use(basePathname ...`. Will fix in a commit after this is merged 😉

ObjectiveTruth: Ah, good catch! and thanks!

PR #181, opened 2017-06-03
Enabling/disabling javascript - fixes #152

### LINKED ISSUE #152: Enabling/disabling javascript
I was testing my website against xss and for fun I tried to sent a mail with some JavaScript inside.
Suddenly, when I clicked on that mail the JavaScript code executed himself. :bowtie:
In production (of course because js is disabled in mails) this is impossible but for testing purpose, I wish you can prevent JS from executing when you click on a mail. 
I think this is contradictory because Mail Dev run on node... 
At least have you any Idea to fix this small glitch?:beetle:

++ Way2key

donut87:
This adds an option to disable the execution of javascript in a mail.

By default Javascript will be executed with the flag --disable-javascript it can be disabled.

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,7 +44,7 @@ module.exports = function (config) {
       .option('-o, --open', 'Open the Web GUI after startup')
       .option('-v, --verbose')
       .option('--silent')
-      .option('-o, --disable-javascript', 'If set, disables the execution of JavaScript in Mails')
+      .option('-j, --disable-javascript', 'If set, disables the execution of JavaScript in Mails')
       .parse(process.argv)
   }
 
--- a/index.js
+++ b/index.js
@@ -44,6 +44,7 @@ module.exports = function (config) {
       .option('-o, --open', 'Open the Web GUI after startup')
       .option('-v, --verbose')
       .option('--silent')
+      .option('-o, --disable-javascript', 'If set, disables the execution of JavaScript in Mails')
       .parse(process.argv)
   }
 
@@ -72,6 +73,10 @@ module.exports = function (config) {
 
   if (config.autoRelay) {
     mailserver.setAutoRelayMode(true, config.autoRelayRules)
+  }
+
+  if (config.disableJavascript) {
+    mailserver.setDisableJavascript()
   }
 
   if (!config.disableWeb) {
--- a/test/middleware.test.js
+++ b/test/middleware.test.js
@@ -100,7 +100,7 @@ describe('middleware', function () {
         .then(function (res) {
           assert.equal(
             res.body,
-            `<img src="//localhost:8080/maildev/email/${email.id}/attachment/tyler.jpg"/>`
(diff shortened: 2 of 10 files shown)


donut87:
I will fix the code coverage soon...

christiandohnert-misterspex:
Ok. Soon is relative :-D
Anyway this should do the trick.

donut87:
As "I" was saying... Soon is relative, but nevertheless I found some time to fix this.

djfarrelly:
Hey @donut87, thanks for this work. I wasn't able to get this to work correctly in my browser

screenshot

I'm wondering if it might be better to use a library like DomPurify on the API side? https://www.npmjs.com/package/dompurify This might make the experience in MailDev more similar to mainstream email clients that do this already!

donut87:
Thanks for looking into this.
I could just say what every developer says: 'It worked on my machine'. I
will have a look into your suggestion. Maybe this might be easier.

donut87:
Finally... This took me far too long. @djfarrelly could you have a look at this? This seems more the way you wanted it, even when this is not the package you suggested.

index.js
```diff
@@ -44,6 +44,7 @@ module.exports = function (config) {
       .option('-o, --open', 'Open the Web GUI after startup')
       .option('-v, --verbose')
       .option('--silent')
+      .option('-o, --disable-javascript', 'If set, disables the execution of JavaScript in Mails')
```
oktapodia: The `-o` option is already used by [--open](https://github.com/djfarrelly/MailDev/pull/181/files#diff-168726dbe96b3ce427e7fedce31bb0bcR44)

test/middleware.test.js
```diff
@@ -100,7 +100,7 @@ describe('middleware', function () {
         .then(function (res) {
           assert.equal(
             res.body,
-            `<img src="//localhost:8080/maildev/email/${email.id}/attachment/tyler.jpg"/>`
+            `<img src="//localhost:8080/maildev/email/${email.id}/attachment/tyler.jpg">`
```
oktapodia: The img HTML tag is no longer valid without the final `/`

donut87:
Oh dear lord. I totally forgot about this pull request. Nevertheless, I'm on it!

index.js
donut87: Changed to '-j'

index.js
```diff
@@ -44,6 +44,7 @@ module.exports = function (config) {
       .option('-o, --open', 'Open the Web GUI after startup')
       .option('-v, --verbose')
       .option('--silent')
+      .option('-j, --disable-javascript', 'If set, disables the execution of JavaScript in Mails')
```
djfarrelly: @oktapodia @donut87, I love this change, but it's got me thinking - should JavaScript be disabled by default as a typical email client would strip out all JS code automatically? This would make this flag something like `-js, --enable-javascript, If set, enables JavaScript execution in emails`

PR #191, opened 2017-08-31
Update Dockerfile

Davidsoff:
This way it is super easy to override the default CLI arguments. 
for example:
`docker run -p 1025:1025 8080:8080 maildev --web 8080 --smtp 1025`


oktapodia:
Awesome! Thanks for your PR.

PR #192, opened 2017-09-02
add lint:fix to package.json

dszymczuk:


oktapodia:
Thanks for your PR but I'm not sure this is necessary

PR #193, opened 2017-09-02
check if files in temp folder exist

dszymczuk:


donut87:
Is there any specific reason to do so? Could you describe the reason for your change?

dszymczuk:
Yes.

I use your mail server in E2E tests. Before each test I use your API to get and delete all mails. And that's point where your app sometimes crashed.
It try to delete mails but it can't because folder / files don't exist.

It could be tricky to reproduce, but you can try send mail and delete all mails 5-10 times in loop.
Maybe you catch error.

dszymczuk:
Any way, this protects your app ;)

oktapodia:
Thanks for your work! Sorry I did not spot your PR earlier, this has already been fixed via the [rimraf](https://www.npmjs.com/package/rimraf) package

PR #196, opened 2017-10-09
Fix version display in GUI

Kerumen:
Usually, we don't display a period between the `v` and the version number.

screen shot 2017-10-09 at 10 33 06


kgeis:
I think we could just get rid of the v altogether.

djfarrelly:
Good catch 😄 Thanks!

PR #202, opened 2018-01-22
Fix #145: Attachments with the same cid

### LINKED ISSUE #145: Attachments overwriting each other
Using 0.14.0, it looks like emails with the same content, and same named attachments are saving the attachments over the top of each other.

I had a brief look and is this due to using `contentId` instead of `generatedFileName` from MailParser's attachment properties? I can check myself when I get a chance, but thought I'd raise it first either way.

GeoffreyBooth:
Fixes #145. Currently, if two emails arrive where both emails use the same `cid` for an attachment, e.g. ``, Maildev overwrites the first attached file with the second one (and so on, if more such emails arrive). This is because Maildev saves each attachment to `tempDir/cid`, where all attachments from all emails share the same folder. This PR changes that behavior so that now attachments are saved in `tempDir/emailId/cid`, so that different emails’ attachments don’t overwrite each other.

To see the bug in action:
```bash
git clone git@github.com:GeoffreyBooth/MailDev.git
cd MailDev

# Check out the commit where I added test data to illustrate the bug,
# but before I committed the fix:
git checkout 82b6e2d80cf5ed46d100ca5b741c357a2c7b7b47

npm install
npm run dev
```

Then go to http://localhost:1080/ and open the messages “Where’s Tyler?” and “Big Wave Coming”. You should see the same photo in both emails, either in the preview or if you click Attachments. Checkout the end of my `attachments` branch (`git checkout attachments`) to see it fixed.

Also includes a test.


GeoffreyBooth:
@djfarrelly?

kgeis:
Thank you for the excellent PR, @GeoffreyBooth. I offered to offload some of the next steps on MailDev from @djfarrelly, and I need to set up a branch with 1.0.0 fixes. I'll try to get that together soon and pull your change to it.

kgeis:
@GeoffreyBooth, could you retarget this PR to the release-1.0 branch?

Issue #166, opened 2017-04-30
Using another mail client of my choosing


akivajgordon:
Is it possible to use another mail client, other than the MailDev web-based client? For example, on macOS, I'd love to use the default Mail.app (or whatever my OS's default client is) in order to view incoming messages. This will enable me to:

* get desktop notifications through Notification Center
* use keyboard shortcuts to navigate
* view full message headers
* not rely on development of the MailDev client, as Mail.app is already full-featured

If this is possible, **can someone share detailed instructions of how to set up the mail account** in System Preferences > Accounts?

If this is not possible, then kindly consider this a feature request.

djfarrelly:
Hey @akivajgordon, this isn't current possible, but it's a clever idea! 

For reading emails, an IMAP module would have to be implemented (example: https://github.com/mscdex/node-imap). A lot of the underlying logic to get this to work already exists ([see the mailsever methods here](https://github.com/djfarrelly/MailDev/blob/master/lib/routes.js)). 

Is this something you'd be interested in writing/contributing?

akivajgordon:
I don't quite have the technical know-how regarding mail servers, but I'll look into it and give it a shot. If I have any questions or updates, or if I give up (😱), I'll post on this thread.

akivajgordon:
@djfarrelly Configuring an account in Apple Mail (and probably other clients as well) requires the following info:

![screen shot 2017-05-12 at 12 47 17 am](https://cloud.githubusercontent.com/assets/6025036/25982691/9831987c-36ac-11e7-8a75-7ff226dce7ad.png)

one of which is an IMAP mail server. You linked to [node-imap](https://github.com/mscdex/node-imap), but isn't that an IMAP client? Would the client just be Apple Mail or Thunderbird or whatever? Seem like the more relevant need is something like [hoodiecrow](https://github.com/andris9/hoodiecrow).

djfarrelly:
Ah, you're right @akivajgordon, that is only a client. We'd have to create an IMAP server, not a client.

akivajgordon:
[Wild Duck](https://github.com/nodemailer/wildduck) looks like a promising option for an IMAP server. And it's built by the same folks who produce [smtp-server](https://github.com/nodemailer/smtp-server), which MailDev is based on.

kgeis:
@akivajgordon, if you download the email source with the extension .eml and open it on a Mac, it will open in Mac Mail by default.

marcosdiez:
[...]

PR #209, opened 2018-04-13
Support for POP3 and IMAP

marcosdiez:
This PR adds `dovecot` to MailDev, which allows it to support POP3 and IMAP

fixes https://github.com/djfarrelly/MailDev/issues/166

One can try it here: https://hub.docker.com/r/marcosdiez/maildev/


oktapodia:
Sorry it's been a long time ago, can you please resolve the conflicts?

clrh:
I like the idea, I would be glad to see a come back of this pr (I unable to do it). I could test.

marcosdiez:
Sorry, but I won't have time to do that. I am in a different project now.
That being said, my container that has both pop and imap still works: marcosdiez/maildev:latest 
that might solve your immediate problem.

andreadelfante:
Hi, any news about this?

djfarrelly:
This is out of the scope of the Maildev project right now. I think this might be better as a standalone package or container - I think this bloats what is already there, although I'm happy to publish a secondary package like `maildev-dovecot`. 

Thanks for this work, if someone wants to work on the above idea, please feel free to comment or propose the solution in a new issue.

PR #224, opened 2018-07-13
Should be able to use both STARTTLS and AUTH (Fixed #187)

### LINKED ISSUE #187: STARTTLS with AUTH
Hi,

I'm running maildev with  `--incoming-user` && `--incoming-pass` but I'm experiencing the following error:
```
SMTP server did not accept STARTTLS with error: 500 Error: command not recognized
```

What is the reason behind disabling the STARTTLS command [right there ?](
https://github.com/djfarrelly/MailDev/blob/3ac3912db4186e722c220050cabd55cd0e1e730b/lib/mailserver.js#L170)

Is it because it's not supported ?

Thank you !

Blufe:
Currently, the case of transmitting both STARTTLS and AUTH can not be covered.

Like smtp-server module, not only extended commands, but all unnecessary commands should be controllable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,7 @@ For convenient use with Grunt, try [grunt-maildev](https://github.com/xavierprio
       --web-pass <password>           HTTP password for GUI
       --base-pathname <path>          base path for URLs
       --disable-web                   Disable the use of the web interface. Useful for unit testing
-      --hide-extensions <extensions>  Comma separated list of SMTP extensions to NOT advertise
-                                      (STARTTLS, SMTPUTF8, PIPELINING, 8BITMIME)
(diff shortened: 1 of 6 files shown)


kgeis:
Your commit comment says 

> Deprecated the '--hide-extensions' option

but in fact it removed the option. I'm a little concerned by the idea of changing command-line options between 1.0.0-rc3 and 1.0.0.

Blufe:
I see.
How do you think that both options coexist and change the new option to take precedence?

oktapodia:
Thanks for your work! 

The release will come without this PR, be sure that we will look into it soon after

oktapodia:
Is it possible for you to improve the tests? I would like to avoid a -6% of code coverage.

PR #237, opened 2018-11-15
Provide a way of limiting the mail storage size

kjagiello:
MailDev may sometimes be used as a part of a development/staging environment where the process may be running for a very long time without interruption. In those cases, it is undesirable to let that
process grow freely. This limit makes sure that the memory usage is bounded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ For convenient use with Grunt, try [grunt-maildev](https://github.com/xavierprio
       --disable-web                   Disable the use of the web interface. Useful for unit testing
       --hide-extensions <extensions>  Comma separated list of SMTP extensions to NOT advertise
                                       (STARTTLS, SMTPUTF8, PIPELINING, 8BITMIME)
+      --store-limit <limit>           Limit the number of stored messages
       -o, --open                      Open the Web GUI after startup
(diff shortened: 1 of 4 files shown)


kgeis:
This is a great feature, and I am looking forward to using it.

Thank you for including a test. Here's what I'd like to see before pulling:
- Update the Usage section of README.md to document the new option.
- Apply the setting with a setter like how autoRelay is done. I see this as a less-used property of mailServer, and I don't want an explosion of parameters on the create(..) method.
- Remove repetition of email count in the test (one instance of 2, two instances of 3, and transporter.sendMail() being called three times).

kjagiello:
@kgeis 

> - Update the Usage section of README.md to document the new option.
> - Apply the setting with a setter like how autoRelay is done. I see this as a less-used property of mailServer, and I don't want an explosion of parameters on the create(..) method.

Good points! Fixed!

> - Remove repetition of email count in the test (one instance of 2, two instances of 3, and transporter.sendMail() being called three times).

I did find some flaws in the test, so I have redone it now. Take a look at it and let me know you see any more issues there.

kjagiello:
@kgeis Do you think we could this merged in sometime soon?

PR #246, opened 2019-03-20
Upgrading to NodeJS 8 LTS version

### LINKED ISSUE #245: Upgrading to the first LTS version
Because the NodeJS v4 is stated as an end of life version, it would be good to upgrade the package to support the first LTS version.

[Source](https://github.com/nodejs/Release)

oktapodia:
Fix #245

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "standard": "10.0.2"
   },
   "engines": {
-    "node": ">=6.0.0"
+    "node": ">=8.0.0"
   },
   "standard": {
     "ignore": [
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "url": "http://github.com/djfarrelly/maildev.git"
   },
   "scripts": {
-    "test": "standard && istanbul cover _mocha",
+    "test": "standard && nyc _mocha --exit",
     "lint": "standard",
     "dev": "node ./scripts/dev.js & npm run css-watch",
     "css": "node-sass --output-style compressed -o app/styles assets/styles/style.scss",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -37,34 +37,45 @@
...
+    "mocha": "^6.0.2",
+    "node-sass": "^4.11.0",
+    "nodemailer": "5.1.1",
+    "nodemon": "^1.18.10",
+    "nyc": "^13.3.0",
+    "standard": "10.0.2"
   },
   "engines": {
-    "node": ">=4.0.0"
+    "node": ">=8.0.0"
```
djfarrelly: 👏

package.json
```diff
@@ -37,34 +37,45 @@
...
+    "async": "2.6.2",
+    "commander": "2.19.0",
+    "express": "4.16.4",
+    "mailparser-mit": "^1.0.0",
+    "opn": "^5.5.0",
+    "rimraf": "^2.6.3",
+    "smtp-connection": "4.0.2",
+    "smtp-server": "3.5.0",
+    "socket.io": "2.2.0",
+    "wildstring": "1.0.9"
```
djfarrelly: Wow, lots of great updates here! @oktapodia Are there any breaking changes w/ major version increases in these libraries, ex. smtp-connection or smtp-server?

oktapodia: I only updated the packages which are not including a major breaking change.

The main changes are:
- Library `opn` instead of `open` (deprecated + contains a vulnerability)
- Library `mailparser-mit` instead of `mailparser` (the latest version is slightly different, the mit version is the same one) 

Another PR will come to update the other packages + some rewrite + optimized tests

PR #247, opened 2019-03-22
Add appveyor configuration

oktapodia:


kgeis:
@oktapodia, I don't understand what issue this resolves. Do you want to use Appveyor instead of Travis or in addition? What is the reason?

oktapodia:
@kgeis only in addition of Travis, I read many issues saying that we have a bug on windows, appveyor is running the tests on a windows machine and will help us to directly spot the introduced bug on windows :)

PR #255, opened 2019-06-17
feat(rest): Add health checks

### LINKED ISSUE #242: Health check for the REST API
When `--web-user` and `--web-pass` options are provided there is no URI we can use for Kubernetes `livenessProbe` and `readinessProbe` checks that returns HTTP 200. Correct? The root `/` as well as every route in the REST API will return a 401 since no auth will be provided by the checks. This prevent us from exposing the web UI when using Kubernetes Ingress. 

Is there a URI we can use with `--web-user` and `--web-pass` options that will return HTTP 200 without provided the auth info? Do you think it would be a good idea to implement a `/healthz` or `/ping` route to the REST API that will provide the above functionality?

loremaps:
Closes #242

### DIFF
--- a/docs/rest.md
+++ b/docs/rest.md
@@ -85,3 +85,5 @@ Returns:
 "to" address.
 
 **GET    /config** - Get the application configuration.
+
+**GET    /healthz** - Health check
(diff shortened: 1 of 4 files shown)


oktapodia:
Thanks for your PR

PR #256, opened 2019-06-17
URI encode attachment file name

andreas:
This PR adds URL encoding of attachment file names. Previously when attached image names contained characters than needed escaping, they would fail to display in HTML view. With this PR, the attachment names are correctly escaped and appear in HTML view.


oktapodia:
Thanks for your PR

PR #261, opened 2019-08-02
Fix typos in var names and logger.

harpocrates12:
Fix typos in var names and logger.


oktapodia:
Good to me! Thank you for your PR

PR #269, opened 2019-10-06
Add an email grouping feature

micoli:
When maildev is used in a CI environment, and multiples tests are run in parrallel, it's rather difficult to clean all emails sent by one test without perturbing other tests.

This feature use a dedicated header in mails to group/categorize them through specific REST routes.
- **GET    /group** - Get all groups, by default, if an email doesn't have the `x-maildev-group` header, then it's group is `default`
- **GET    /group/:group** - Get all emails by group
- **DELETE /group/:group** - Delete all emails by group


oktapodia:
Sounds like a very good idea! Can you please add some tests to your pull request?

micoli:
I added one API test. It sounds more like an integration test, It test : 
- groups listing
- mail by group 
- group listing after mail deletion by group.

micoli:
I have just rebased thoses modifications, included the added tests. Could you tell me what you think about it ?

micoli:
as a complement, the changes on the UI below:
image

micoli:
PR is rebased

Issue #244, opened 2019-03-19
Change default Dockerfile port from 80/25


djfarrelly:
### Purpose
I think it would better for Docker container security to not run the main command on a port under 1000. 

### Idea
Change the Dockerfile to run on 1080 / 1025 instead of 80 / 25. The challenge is that this could break people's usage when upgrading so we'd have to think about this one or save it for a 2.x breaking changes release.

bearcodi:
Hi @djfarrelly I've created a pull request for this, I kept getting errors when using the docker image on Mac when I was trying to bind to port `25`

The pull request sets the default ports for web and smtp to 1080 and 1024 respectively.

Thanks for the cool project! Its in my dev/test toolkit by default!

PR #273, opened 2019-11-04
Updated default web and smtp ports

bearcodi:
Hi

I've updated the default ports for the web and smtp arguments.

Web now runs on port 1080 and SMTP runs on port 1025.

Proposed fix for #244


djfarrelly:
Thanks for your contribution @bearcodi! This change was duplicated as part of the larger changes in #281. I'm going to close this now, but we really appreciate your PR!

PR #280, opened 2019-12-18
Add option to specify directory for mail storage

arminfriedl:
Mails and attachments are not stored in a os dependent temporary directory, but
in a given directory. This is useful in containerized environments where
persistent volumes are mapped to folders in the container.

### DIFF
--- a/lib/mailserver.js
+++ b/lib/mailserver.js
@@ -19,7 +19,7 @@ const outgoing = require('./outgoing')
 const defaultPort = 1025
 const defaultHost = '0.0.0.0'
 const store = []
-const tempDir = path.join(os.tmpdir(), 'maildev', process.pid.toString())
+const defaultMailDir = path.join(os.tmpdir(), 'maildev', process.pid.toString())
 const eventEmitter = new events.EventEmitter()
 
 /**
@@ -54,7 +54,7 @@ function saveEmail (id, envelope, mailObject) {
   object.time = new Date()
   object.read = false
   object.envelope = envelope
-  object.source = path.join(tempDir, id + '.eml')
+  object.source = path.join(mailServer.mailDir, id + '.eml')
 
   store.push(object)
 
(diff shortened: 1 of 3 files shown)


oktapodia:
Thank you for your PR, can you please add tests?

straurob:
I implemented the same pull request as I found this one is already existing :-)

Is there anything I can contribute to get this feature merged? We need this feature in our current project.

lib/mailserver.js
```diff
@@ -113,47 +113,46 @@ function handleDataStream (stream, session, callback) {
 }
 
 /**
- * Delete everything in the temp folder
+ * Delete everything in the mail directory
  */
-
-function clearTempFolder () {
-  fs.readdir(tempDir, function (err, files) {
+function clearMailDir () {
```
audig, line 118: This method will always delete at startup the directory so we loose the feature to get back mail if a container restart

audig:
Inspired from your work, but with some changes : https://github.com/cnieg/maildev/commit/939b5336a7fff343d0514764842e5cd69e72a80e

- Add a method to load mail in the store from directory
- Add a route to reload mail from directory. We can have a cron to purge some outdated mails and have to reload mails (ie helm charts for kubernetes : https://github.com/cnieg/helm-charts/blob/maildev/charts/maildev/values.yaml#L37)
- Disable clean of data dir at startup

arminfriedl:
> Inspired from your work, but with some changes : [cnieg@939b533](https://github.com/cnieg/maildev/commit/939b5336a7fff343d0514764842e5cd69e72a80e)
> 
>     * Add a method to load mail in the store from directory
> 
>     * Add a route to reload mail from directory. We can have a cron to purge some outdated mails and have to reload mails (ie helm charts for kubernetes : https://github.com/cnieg/helm-charts/blob/maildev/charts/maildev/values.yaml#L37)
> 
>     * Disable clean of data dir at startup

Sorry for having no time to follow up on this. Thank you for your PR @audig I will close this one then.

lib/mailserver.js
arminfriedl: Yes, I didn't change this, as this behaviour was no issue for me. The persistent volume was needed because of storage restrictions on the container/pod. MailDev was used as a transient (but potentially high volume) testing environment. But you are right that if we already use a persistent volume, it is probably the expected behaviour that this does not get cleared out on restart.

arminfriedl:
Closing in favour of #303

PR #281, opened 2020-01-12
Remove root user from docker, update ports and update Node / libraries

### LINKED ISSUE #244: Change default Dockerfile port from 80/25

jovanmaric:
This PR adds the following:

- **Change the default user from root to node in image**
  Currently the user is set as root, which is unsafe. I have changed the user to the default node user, 
  along with the working directory to /home/node, so that the dependencies and the app are owned by  the node user. Keep in mind that the node user can still overlap because of its id and group id (1000, 1000 respectively).  
- **Increase the default ports with 1000:**
  Ports 80 and 25 are mostly reserved for other containers / host applications. These are changed to 
  1080 and 1025, which better isolates this application for development. This fixes #244 and makes 
  #273 obsolete. This is a needed fix for podman users as well, if ports '< 1024' aren't enabled in the 
  user space.
- **Change the entrypoint to an absolute path:**
  Latest Podman is having trouble with relative paths for some reason. This makes the binary location more explicit as well.
- **Updated node to the latest LTS:**
  I'm not sure what implications this has for users which aren't using containers. Can this application 
  be included as a library? I have made the assumption here that it is meant to be run standalone / 
  containerized, and therefore bumping the version to the latest LTS should be fine. I could change 
  this to V10(Active LTS until 2021) if the version is too high?
- **Updated libraries and fixed npm audit security issues:**
  Fixed the 75 security issues which npm audit reported. 

I wondering what you think of this PR. It would be great to have a version bump on dockerhub as well if this gets merged.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,8 @@ ENV NODE_ENV production
 
 # Build
 FROM base as build
-WORKDIR /root
 
+WORKDIR /root
 COPY package* ./
 
 RUN apk add --no-cache curl \
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ USER node
 
 WORKDIR /home/node
 
-ADD . /home/node
+COPY --chown=node:node . /home/node
 
 RUN npm install && \
     npm prune && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,18 +3,22 @@ MAINTAINER "Dan Farrelly <daniel.j.farrelly@gmail.com>"
 
 ENV NODE_ENV production
 
-RUN apk add --no-cache curl
+WORKDIR /root
 
-USER node
+COPY package* .
+
+RUN apk add --no-cache curl \
+  && npm install \
+  && npm prune \
+  && npm cache clean --force \
+  && rm package*
+
+COPY . /home/node
+RUN mv /root/node_modules /home/node \
+    && chown -R node:node /home/node
 
 WORKDIR /home/node
-
-COPY --chown=node:node . /home/node
-
-RUN npm install && \
-    npm prune && \
-    npm cache clean --force && \
-    rm -rf /tmp/*
+USER node
 
 EXPOSE 1080 1025
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,12 @@
-FROM node:12-alpine
+# Build
+FROM node:10-alpine as build
 MAINTAINER "Dan Farrelly <daniel.j.farrelly@gmail.com>"
 
 ENV NODE_ENV production
 
 WORKDIR /root
 
-COPY package* .
+COPY package* ./
 
 RUN apk add --no-cache curl \
   && npm install \
@@ -13,17 +14,19 @@ RUN apk add --no-cache curl \
   && npm cache clean --force \
   && rm package*
 
-COPY . /home/node
-RUN mv /root/node_modules /home/node \
-    && chown -R node:node /home/node
+# Prod
+FROM node:10-alpine
 
-WORKDIR /home/node
 USER node
+WORKDIR /home/node
+
+COPY --chown=node:node . /home/node
+COPY --chown=node:node --from=build /root/node_modules /home/node/node_modules
 
 EXPOSE 1080 1025
 
-ENTRYPOINT ["/home/node/bin/maildev"]
-CMD ["--web", "1080", "--smtp", "1025"]
-
-HEALTHCHECK --interval=10s --timeout=1s \
-  CMD curl -k -f -v http://localhost:1080/healthz || exit 1
+# ENTRYPOINT ["/home/node/bin/maildev"]
+# CMD ["--web", "1080", "--smtp", "1025"]
+#
+# HEALTHCHECK --interval=10s --timeout=1s \
+#   CMD curl -k -f -v http://localhost:1080/healthz || exit 1
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,13 +8,13 @@ ENV NODE_ENV production
 FROM base as build
 
 WORKDIR /root
-COPY package* ./
+COPY package*.json ./
 
 RUN apk add --no-cache curl \
   && npm install \
   && npm prune \
   && npm cache clean --force \
-  && rm package*
+  && rm package*.json
 
 # Prod
 FROM base as prod
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,25 @@
-FROM node:8-alpine
+FROM node:12-alpine
 MAINTAINER "Dan Farrelly <daniel.j.farrelly@gmail.com>"
 
 ENV NODE_ENV production
 
 RUN apk add --no-cache curl
 
-RUN mkdir -p /usr/src/app
-WORKDIR /usr/src/app
+USER node
 
-ADD package*.json /usr/src/app/
+WORKDIR /home/node
+
+ADD . /home/node
 
 RUN npm install && \
     npm prune && \
     npm cache clean --force && \
     rm -rf /tmp/*
 
-ADD . /usr/src/app/
+EXPOSE 1080 1025
 
-EXPOSE 80 25
-
-ENTRYPOINT ["bin/maildev"]
-CMD ["--web", "80", "--smtp", "25"]
+ENTRYPOINT ["/home/node/bin/maildev"]
+CMD ["--web", "1080", "--smtp", "1025"]
 
 HEALTHCHECK --interval=10s --timeout=1s \
   CMD curl -k -f -v http://localhost/healthz || exit 1
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,11 @@
-# Build
-FROM node:10-alpine as build
+# Base
+FROM node:10-alpine as base
 MAINTAINER "Dan Farrelly <daniel.j.farrelly@gmail.com>"
 
 ENV NODE_ENV production
 
+# Build
+FROM base as build
 WORKDIR /root
 
 COPY package* ./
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,26 +1,25 @@
...
 ENV NODE_ENV production
 
 RUN apk add --no-cache curl
 
-RUN mkdir -p /usr/src/app
-WORKDIR /usr/src/app
+USER node
 
-ADD package*.json /usr/src/app/
+WORKDIR /home/node
```
jderusse: what's the benefit of hosting the application in a non-standard path?

Dockerfile
```diff
@@ -1,26 +1,25 @@
...
 
 ENV NODE_ENV production
 
 RUN apk add --no-cache curl
 
-RUN mkdir -p /usr/src/app
-WORKDIR /usr/src/app
+USER node
 
-ADD package*.json /usr/src/app/
```
jderusse: adding package.json in separate layer reduces build time.

Dockerfile
```diff
@@ -1,26 +1,25 @@
...
 
 RUN npm install && \
     npm prune && \
     npm cache clean --force && \
     rm -rf /tmp/*
 
-ADD . /usr/src/app/
-
-EXPOSE 80 25
+EXPOSE 1080 1025
```
jderusse: I don't get the benefit of not-running the app on standard port?

#244 mention security issue without telling what the issue is.

jovanmaric: Well, it's general convention to not publishing a dev container to the world. Considering that this is used in dev environments, it should be restricted to a port which isn't by default open. All distro firewalls have port 80 opened by default, otherwise http(s) traffic would be blocked by default. This isn't a problem with expose, because expose is used to link containers with a network created by docker itself. However, considering that the maildev server is running in broadcast mode (0.0.0.0), this would mean that every time that a user wants to view the maildev webpage, the user would publish the container to the world. Having these ports set on to higher port values would allow firewalls to block this by default.

Another point could be made regarding that port sniffers would try the default used ports first. Port 80 in combination with basic firewall settings would be really easily found, considering that port 80 is the default webserving port.

The last point is regarding podman. Currently fedora supports running rootless containers by default, and this is becoming a standard really quickly. Ports below 1024 aren't available to users, without setting a kernel flag. This change would therefore make maildev future proof.

jderusse:
not using the default port is a big :-1: for this PR 

huge drawback for small benefit

Dockerfile
jovanmaric: I have reverted caching. No idea why I thought the amount of packages was low :confused:..

Dockerfile
jovanmaric: Because /usr/src is meant for kernel and header files. Not applications. Also having the application in the /home/node path, brings it into the node userspace. Node packages should never have root rights, to minimize chances of privilege escalation.

Dockerfile
jderusse: your assumption are wrong. You don't need to expose container port to host machine, and if you do it, you don't have to use the same port in your mapping.

This is exactly what is performed with nginx or databases in Dev environment.

That's a huge drawback just for Fedora users (sI really wonder how they run nginx/mysql/...). 

I suggest to watch other distribution (if they restrict root users too) and other Images that expose port <1024 before applying such change.

Dockerfile
jderusse: home is for user, not application/services

Dockerfile
jderusse: what's about https://medium.com/elbstack/docker-go-and-privileged-ports-d6354db472c3 to use standard port with non root user?.

jovanmaric: > your assumption are wrong. You don't need to expose container port to host machine, and if you do it, you don't have to use the same port in your mapping.

Sure, you can remap the ports however you like. However a sensible safe default image is better than having to retain to custom configuration. These changes are meant to make sure  that beginners and/or overlooked mistakes don't get you exposed for no reason with default docker conventions. 

> This is exactly what is performed with nginx or databases in Dev environment.

In contrast to containers, nginx doesn't work without configuration. You can't start the nginx service and expose your dev applications to the internet, without actually having to think about how you are exposing it. 

> That's a huge drawback just for Fedora users (sI really wonder how they run nginx/mysql/...).

Well you can just set `net.ipv4.ip_unprivileged_port_start=0` and allow users to use privileged ports. It isn't just in reference to fedora users, k8s and docker can currently run in some form rootless, lxc as well. Podman on fedora is just an example. Running podman rootless on arch myself without any hassle.

Running rootless nginx and mysql containerized isn't any different than running them as root. Also keep in mind that the default mysql port is 3306, therefore working without any kernel flags.

> I suggest to watch other distribution (if they restrict root users too) and other Images that expose port <1024 before applying such change.

I'm not sure what you're trying to say. The PR changes have nothing to do with restricting root users and privileged ports. If you mean that I should check other maildev like apps to get inspired. Well...
- https://github.com/mailhog/MailHog/blob/master/Dockerfile
- https://github.com/sj26/mailcatcher/blob/master/Dockerfile

Dockerfile
jovanmaric: So what do you propose? /home/node should be fine for the application. Either way I'll have to chown the app dir to keep it usable as node user.

Dockerfile
jovanmaric: > what's about https://medium.com/elbstack/docker-go-and-privileged-ports-d6354db472c3 to use standard port with non root user?.

Are you seriously trying to teach me how to expose the container to 80/25, just so you can keep using the ports? :rofl:

jderusse: this conversation is meaningless

Talking about using a dev tool in K8s, exposing the container network to the Host machine, ignoring Images that expose 25/80, being arrogant.

jovanmaric: > Talking about using a dev tool in K8s

I'm not talking about any dev tool?

> exposing the container network to the Host machine

I'm aware of the ambiguity, but I was referring to the act of exposing the container to the world.

> ignoring Images that expose 25/80

What do you mean?

> being arrogant

The last comment was meant to be playful.. But anyway,

The only way I can see port 80 being a security issue is that beginner- or configuration mistakes are made, as I have tried to make clear in my above comments (I have edited them a bit so they are more clear regarding this). Expose is meant as a type of documentation, and the Dockerfile is mostly the first reference of needed ports when executing `docker run --publish x:y ...`. Having this set to a sensible default in combination with running the container as the node user, would make it more secure for beginners and it overlaps perfectly with default rootless container setups.

Dockerfile
```diff
@@ -1,26 +1,29 @@
...
 MAINTAINER "Dan Farrelly <daniel.j.farrelly@gmail.com>"
 
 ENV NODE_ENV production
 
-RUN apk add --no-cache curl
+WORKDIR /root
 
-RUN mkdir -p /usr/src/app
-WORKDIR /usr/src/app
+COPY package* .
```
djfarrelly: This fails with the error:
> When using COPY with more than one source file, the destination must be a directory and end with a /

_Docker version 19.03.5_

Can you update this to instead copy to `/root/`?

Dockerfile
djfarrelly: Changing from the root user to a `node` user and using a port >1000 apply the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege). It's a smart best practice to follow and as this is open source software, we should default to the least priviledge. If someone cares to run their own images as root, that's fine, but let's have good defaults for the most broad usage.

package.json
```diff
@@ -55,15 +55,15 @@
...
+    "mocha": "^6.2.2",
+    "node-sass": "^4.13.0",
+    "nodemailer": "^6.4.2",
+    "nodemon": "^1.19.4",
     "nyc": "^14.1.1",
     "standard": "^12.0.1"
   },
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=12.0.0"
```
djfarrelly: > I could change this to V10(Active LTS until 2021) if the version is too high?
Yes, @jovanmaric, can you change this to 10? With 10 in LTS, I don't want users running 10 to get an error w/ this new min engine requirement unless there is another core dependency in 12 that maildev requires. Thanks!

jovanmaric:
> Thanks for your contribution @jovanmaric! I left a couple comments to request changes. One other idea that we could switch to for the Dockerfile is to use a [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) which would allow the dependency install and install in the `/root/` dir to be separate from the final image, also reducing layers of our final image.

Yes I agree, multi-stage builds are cleaner and more elaborative, however only the targeted stage is actually reduced in layers. The ci system would still need to cache all the needed stages to not require a rebuild every time. Also this would require only two stages, as dividing the stages into "development" and "production" only really works with the new buildkit in docker. Otherwise all stages will be automatically built :sweat_smile:. Though this last bit is more of an optimization. 
 
> Would love to hear your thoughts on that too, although, once these couple changes are made, we can merge this in and publish. I'm just now debating if this change should mean a 2.0 upgrade as it's a breaking change for the Docker image.

Maybe a 2.0 upgrade would be a better fit yes, as I can't tell if this is used in production like environments like staging or review apps. 

Regarding your other comments, ill take a look later this evening.

jovanmaric:
>  Thanks for incorporating the changes I requested as well as your response re:version @jovanmaric!
> 
> I'll merge this today and starting thinking through what the next release will look like as there are lots of other good useful changes in open PRs!

I'm glad this got merged, thanks! :). Feel free to ping me if you need help with any other issue's / PR's. I'm not that experienced with js / node (other than vue/nuxt and some random dabbles in node), but can always help as a sparring partner.

Cheer!

PR #283, opened 2020-02-08
remember autoShow state in locale storage (if available)

frankdekker:
The autoShow state wasn't remember between page refreshes.

When MailDev interface loads the autoShow state will be retrieved from local storage.
When autoShow state changes, the state will be persisted in local storage.


matthijsthoolen:
@oktapodia any chance you can take a look at this pull request?

djfarrelly:
Thanks for your contribution @frankdekker! With your other contribution in #284, I decided to use this idea and expand it to both app settings :)

PR #284, opened 2020-02-08
Browser notifications

frankdekker:
Added notification support to the Web Interface.
- icon added to enable and request notification permissions
- throttle added when notifications are enabled, to only send once every 2 seconds

As Chrome/Firefox require secure context to allow notifications, added https cli arguments
See docs/https.md for details.


matthijsthoolen:
@oktapodia and maybe also this one, we would like this feature :)

app/scripts/controllers/main.js
```diff
@@ -3,13 +3,17 @@
...
+var refreshTimeout = null
+var notificationTimeout = null
 
 app.controller('MainCtrl', [
   '$scope', '$rootScope', '$http', 'Email', '$route', '$location', 'Favicon',
   function ($scope, $rootScope, $http, Email, $route, $location, Favicon) {
     $scope.items = []
     $scope.configOpen = false
     $scope.currentItemId = null
+    $scope.notificationsSupported = 'Notification' in window
```
djfarrelly: I think we should give user feedback if it's supported also using `isSecureContext`.

app/scripts/controllers/main.js
```diff
@@ -93,6 +108,27 @@ app.controller('MainCtrl', [
       $scope.autoShow = !$scope.autoShow
     }
 
+    $scope.enableNotifications = function () {
+      if (window.Notification && window.Notification.permission === 'granted') {
+        window.alert('To disable notifications, revoke the permissions in your browser.')
```
djfarrelly: Instead of this alert, can we instead set `$scope.webNotifications = false`? This context will of course be lost when the browser window is closed, but it might be a better workflow. Perhaps if this was coupled with some setting storage in `localStorage` that may provide the best experience.

index.js
```diff
@@ -76,9 +79,15 @@ module.exports = function (config) {
   }
 
   if (!config.disableWeb) {
+    const secure = {
```
djfarrelly: I like this grouping of the options together. Makes me think that `web.start()`'s arguments should be refactored soon.

app/scripts/controllers/main.js
```diff
@@ -93,6 +108,27 @@ app.controller('MainCtrl', [
...
+        return
+      }
+      window.Notification.requestPermission().then(function (permissions) {
+        $scope.webNotifications = permissions === 'granted'
+      }).catch(function () {
+        window.alert('Unable to enable web notifications. See console for more information')
+      })
+      if (!window.isSecureContext && window.console) {
+        console.info(
```
djfarrelly: I'd rather make this more obvious to the user, perhaps using window.alert or a `title` on the html element.

docs/https.md
```diff
@@ -0,0 +1,32 @@
+# HTTPS
+By default MailDev will run on the http protocol. For Web Notification support you'll need
```
djfarrelly: One can also use web notification support on localhost in most browsers. We should update this copy.

docs/https.md
```diff
@@ -0,0 +1,32 @@
...
+Add the following arguments to your MailDev startup:
+```shell script
+--https
+--https-key /etc/ssl/private/maildev.key
+--https-cert /etc/ssl/certs/maildev.crt
+```
+
+## Open maildev with https
+```
+https:\\192.168.1.103:1080
```
djfarrelly:
```suggestion
https:\\192.168.1.103:1080
https://192.168.1.103:1080
```

app/scripts/controllers/main.js
frankdekker: Interesting, hadn't thought about that. So instead of revoking permissions, you just silence the notifications. Great idea. The silence setting has to be stored preferrably. I as user found it already annoying to toggle the "autoShow" each time.

app/scripts/controllers/main.js
frankdekker: Interesting, what kind of feedback are you thinking of?

docs/https.md
frankdekker: Ah. I wasn't aware localhost was the exception. Indeed useful to be added.

app/scripts/controllers/main.js
djfarrelly: Thanks for the input - all settings will be persisted after #291 :)

app/scripts/controllers/main.js
djfarrelly: I decided to add a `title` attribute on the Notifications toggle now. Also with the new toggle design, Notifications is now greyed out if the browser context does not support it:

Screen Shot 2020-03-30 at 9 18 57 AM

PR #305, opened 2020-05-25
Enable compression when fetching all mails

leonardobsjr:
Since we use maildev as a central mail development server, the number of emails pile really fast and because most of us are WFH, the /email route was taking too long to execute and because there's no indication that maildev is still fetching, some developers thought that the emails had vanished. 

Because of this, we decided to compress the /email route for faster fetching.


soulteary:
Thanks for the feedback, I will review the code and give you feedback later.

There are some conflicts in the current code, I will deal with it in fork first https://github.com/soulteary/maildev/pull/7/

If everything works, your name will appear in the v2.0.2 version, as a contributor, please continue to submit and play together @leonardobsjr

PR #308, opened 2020-06-10
Setting autorelay email only if it is valid.

sumeete:
Background:
-I have wanted to use maildev and set all the options using env vars. The issue I ran into was setting the autorelay to be enabled for all destinations via the env var `MAILDEV_AUTO_RELAY`.
It only sets auto relay to be enabled if I set it, which is fine. However if I set it to something like "1" it will try to forward the email to "1". Rather than changing all the options this will allow you to specify an email to relay to and if you don't specify an email and just fill the var then it will trigger it to auto relay to all destinations. 

Happy to work on this to make it more customize-able but this was the easiest non-breaking change to do for the time being.


sumeete:
Also in a subsequent diff I want to make a change to the dockerfile to remove the `ARGS` sections. For some reason this conflicts with reading envars in docker (reproducible) and if you remove it you can still specify a custom ARGS anyway.

PR #309, opened 2020-06-11
Docker compose envvar fix

sumeete:


Dockerfile
```diff
@@ -28,7 +28,6 @@ COPY --chown=node:node --from=build /root/node_modules /home/node/node_modules
 EXPOSE 1080 1025
 
 ENTRYPOINT ["/home/node/bin/maildev"]
-CMD ["--web", "1080", "--smtp", "1025"]
```
sumeete, line 31: for some reason these can't bereplaced  in a docker compose file.

PR #311, opened 2020-06-20
docs: adjust document wording

bish0polis:
Like Traffic and Knowledge, the plural of Mail and Email is still Mail and Email.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![NPM Version](https://img.shields.io/npm/v/maildev.svg)](https://www.npmjs.com/package/maildev)
 [![JavaScript Style Guide](https://img.shields.io/badge/code_style-standard-brightgreen.svg)](https://standardjs.com)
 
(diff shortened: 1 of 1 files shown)


soulteary:
@LoneRifle @djfarrelly For this detail, which word do you think would be more appropriate?

LoneRifle (approved):
imo, the term "email" and "emails" can be used, depending on context. If referring to more than one email that is still a tractable number, "emails" should be used (as in `5 unread emails`). In this context, email can be used as a non-counting noun.

bish0polis:
> imo

There are rules for that kind of word independent of opinion.  Don't confuse 'syntactically correct' with 'popular'.

soulteary:
> I think this does read better than before, so along with @LoneRifle - I'm for this change :)

OK, then I'll merge this pr, and deal with conflicts. thanks @bish0polis @LoneRifle @djfarrelly

soulteary:
Thanks for the contribution, the code has been merged, in the next version, we will add you as a contributor, looking forward to your more participation. @bish0polis 

https://github.com/maildev/maildev/pull/395

PR #321, opened 2020-09-11
Support clustered processes

courtney-regis:


courtney-regis:
Apologies for the multiple commits but did this on mobile so didn't have the luxury of an IDE

soulteary:
Thanks for the feedback, it looks like the feature is not fully implemented.

In the docker era, multiple processes can be implemented through docker containers, but for users that do not use docker, it may be a good choice to add an option to allow opening and setting the specified number of processes.

Looking forward to your follow-up submissions. @courtney-regis

PR #322, opened 2020-10-01
add dark theme

bazaim:
Create a dark theme.
Add a toggle on the toolbar to toggle the dark theme.
Save setting `darkThemeEnabled` into LocalStorage.


DominikSerafin:
Hey — I've merged this (with some changes) into my [fork](https://github.com/DominikSerafin/maildev) — more info here https://github.com/maildev/maildev/issues/335.

app/index.html
```diff
@@ -117,6 +117,22 @@
...
+                        <li>
+                          <label
+                            class="toggle-switch light"
+                            for="toggle-dark-theme"
+                            title="Enable dark theme">
+                            <input
+                              type="checkbox"
+                              name="theme"
+                              id="toggle-dark-thme"
+                              ng-checked="settings.darkThemeEnabled"
```
benatshippabo, line 130: suggestion: we can skip maintaining the dark theme toggle state and just wrap your dark theme css styles with the `prefer-color-cheme` media query https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme. That way, it will change depending on if the user has their OS set to a dark theme.

assets/styles/_dark-theme.scss
```diff
@@ -0,0 +1,106 @@
+body.dark-theme {
+
+  $dark: darken($brand, 40);
```
benatshippabo, line 3: suggestion: deep black `#03070b`  tends actually strain your eyes more. Example would be the MacOS dark theme, which is considered a pretty good dark theme, versus the Windows 10 dark theme implementation, which is not so good. Maybe use a lighter dark gray instead similar to what the MacOS mail app uses `#1E1E1E` for backgrounds.

app/index.html
bazaim: Good idea, I'll do it to set the default value of `settings.darkThemeEnabled`.  
I'll leave the toggle here, because some mails are unreadable with the dark mode.

assets/styles/_dark-theme.scss
bazaim: Thank you for he feedback, I'll change it ASAP and keep it in mind.

DominikSerafin:
Merging to incorporate further changes from my [fork](https://github.com/DominikSerafin/maildev) that depend on this PR.

PR #323, opened 2020-10-04
feat: Add email embedded favicon supports

micoli:
Add the support of email favicon. They are displayed in the item list on the left of subject and recipient addresses. It's using colorscheme to keep same address email with same color.

image


scripts/send.js
```diff
@@ -48,7 +48,9 @@ var messages = [
     to: 'Johnny Utah <johnny.utah@fbi.gov>',
     subject: 'The ultimate price',
     text: 'If you want the ultimate, you\'ve got to be willing to pay the ultimate price. \nIt\'s not tragic to die doing what you love.',
-    html: '<!DOCTYPE html><html><head></head><body style="background:#eeefont-family:sans-serifpadding:2em 2em">' +
+    html: '<!DOCTYPE html><html><head>' +
+          '<link rel="Shortcut Icon" type="image/x-icon" href="http://favicons.fr/icons_/105b697f95b9749509401e5b6eea24c0_preview_favicon-16.png" />' +
```
soulteary, line 52: There is a need to use local resources instead of online resources

package.json
```diff
@@ -44,6 +44,7 @@
     "cors": "^2.8.5",
     "express": "^4.17.1",
     "mailparser-mit": "^1.0.0",
+    "node-html-parser": "^1.2.21",
```
soulteary, line 47: This package doesn't seem to be used

scripts/send.js
soulteary, line 52: There is a need to use local resources instead of online resources

app/styles/style.css
```diff
@@ -1,4 +1,4 @@
...
...:max-height 500ms ease-in-out}.slide-container.open{max-height:1000px}.code{background-color:#fafafa;border:1px solid #dcdcdc;padding:1em 1em}code{background-color:#fafafa;border:1px solid #dcdcdc;padding:0.1em 0.3em}.search-container{position:relative;padding:0.3em 0.8em 0}.search-icon{position:absolute;top:0.7em;left:1.3em;color:rgba(255,255,255,0.8)}.search-input{display:block;margin:0;padding:0.3em 0.2em;width:98%;text-indent:1.9em;background-color:#4994ce;border:0;color:#fff;-webkit-font-smoothing:antialiased}.search-input:focus{outline:none;background-color:#307ab4}.search-input:focus+.search-icon{color:#fff}::-webkit-input-placeholder{-webkit-font-smoothing:antialiased;color:rgba(255,255,255,0.8);-webkit-transition:color 250ms linear;transition:color 250ms linear}::-moz-placeholder{color:rgba(255,255,255,0.8);-moz-transition:color 250ms linear;transition:color 250ms linear}::-ms-placeholder,::placeholder{-webkit-font-smoothing:antialiased;color:rgba(255,255,255,0.8);transition:color 250ms linear}.preview-iframe{border:0;width:100%;min-height:600px;position:relative}.email-item.current{background:#eaf2f9}.autoshow-toggle{float:right;cursor:pointer;color:white;padding-top:3px}
```
soulteary, line 4: We shouldn't need so many icons

app/scripts/services.js
```diff
@@ -12,3 +12,70 @@ app.service('Email', ['$resource', function ($resource) {
...
+      if (assignedPalette[i].addresses.indexOf(address) !== -1) {
+        return assignedPalette[i].color;
+      }
+    }
+    lastAssigned = (lastAssigned + 1) % assignedPalette.length
+    assignedPalette[lastAssigned].addresses.push(address)
+    return assignedPalette[lastAssigned].color;
+  }
+  /*console.log([
+  assignColor('a1.t@email.com'),
```
soulteary, line 48: Unnecessary comments should be removed

app/components/color-scheme.js
```diff
@@ -0,0 +1 @@
...return this.saturation[t]=e,this.value[t]=n},e.prototype.set_variant_preset=function(t){var e,n,r;for(r=[],e=n=0;n<=3;e=++n)r.push(this.set_variant(e,t[2*e],t[2*e+1]));return r},e.prototype.get_hex=function(t,e){var n,r,o,a,i,u,s,h,l,c,f,p,_,m,d,v;for(l=Math.max.apply(Math,function(){var t,e,n,o;for(n=["red","green","blue"],o=[],e=0,t=n.length;e<t;e++)r=n[e],o.push(this["base_"+r]);return o}.call(this)),Math.min.apply(Math,function(){var t,e,n,o;for(n=["red","green","blue"],o=[],e=0,t=n.length;e<t;e++)r=n[e],o.push(this["base_"+r]);return o}.call(this)),v=255*(e<0?this.base_value:this.get_value(e)),m=e<0?this.base_saturation:this.get_saturation(e),i=l>0?v/l:0,p=[],f=["red","green","blue"],h=0,u=f.length;h<u;h++)r=f[h],_=Math.min.apply(Math,[255,Math.round(v-(v-this["base_"+r]*i)*m)]),p.push(_);for(t&&(p=function(){var t,e,r;for(r=[],e=0,t=p.length;e<t;e++)n=p[e],r.push(51*Math.round(n/51));return r}()),o="",c=0,s=p.length;c<s;c++)a=p[c],d=a.toString(16),d.length<2&&(d="0"+d),o+=d;return o},e}(),t}(),"undefined"!=typeof module&&null!==module&&null!=module.exports?module.exports=t:"function"==typeof define&&define.amd?define([],function(){return t}):window.ColorScheme=t}).call(this);
```
soulteary, line 1: Perhaps the uncompressed code should be committed

soulteary (changes requested):
The current version of the code is not suitable for merging into the repository

soulteary:
Thanks for the feedback, this is an interesting feature, I mentioned some issues in the code review, if you can fix it, I think we should see this feature in the next version.

And note, that we've released a new version and the conflicts contained in this commit need to be resolved as well.

Looking forward to your continued participation @micoli

package.json
micoli: done, i've removed it

app/components/color-scheme.js
micoli: done, i've added an `update-component` rules to put them in components

scripts/send.js
micoli: Some emails get a type="image/x-icon" in their body. half of that PR is to deiplay those icons. I don't see any other way to demonstrate that feature in thoses examples.

app/styles/style.css
micoli: I just used `yarn css` to generate css. those icons were included by that command. Is there is any other way to generate the css file ?

PR #328, opened 2020-11-25
Corrected port mapping for SMTP & Web GUI in docker run command

ravirdv:
Hi,

Looks like port mapping are incorrect in the docker command as the service runs on port 25 & 80 as default on the container. I've changed it to 25 for SMTP service and 80 for WebGUI.

Thanks.


soulteary:
@ravirdv  Thanks for the feedback, the project has fixed this problem https://github.com/maildev/maildev/issues/382 , welcome to join us to continue to improve this project. https://github.com/maildev/maildev/issues/379

Issue #330, opened 2020-12-03
Retrieve emails by subject - RestAPI


1dson:
From the README it's not that clear but is it possible to retrieve an email by subject using the REST API?

djfarrelly:
Yes — you can filter by subject via the REST API using a query param on the list endpoint:

```
GET /api/email?subject=<exact subject>
```

Multiple params are ANDed, and you can also filter by other fields (e.g. `to.address`, `from.address`). See the `/api/email` handler in `packages/api/src/server.ts`.

One caveat: matching is **exact string equality**, not substring/partial. For fuzzy/substring subject search, the MCP layer's `maildev_search_emails` tool does that. Closing as answered — if you'd like substring subject matching in the REST endpoint specifically, please open a fresh issue for it.

PR #334, opened 2020-12-21
feat: allow filtering of emails in http endpoint

madflow:
* This PR allows simple filtering in the HTTP endpoint `/email`
* Use case is - when using in CI (Cypress E2E), it can be useful to only fetch specific emails. Also #330 
* You can only find emails, which correspond exactly to the query parameters.
* Nested fields in the payload can be addressed with a (`to.name`, `headers.x-some-header`) dot syntax

`http http://localhost:1080/email\?to.name\="Johnny Utah"\&subject\="The ex-presidents are surfers"`


kowalczyka:
Hello,
does this change allows filtering only for endpoint in REST API or does it also affects Web UI?

madflow:
> Hello,
> does this change allows filtering only for endpoint in REST API or does it also affects Web UI?

Only for simple filtering in the HTTP endpoint /email.

test/utils.test.js
```diff
@@ -14,4 +14,20 @@ describe('utils', () => {
...
   })
+
+  describe('filterEmails', () => {
+    const emails = [
+      { subject: 'Test', headers: { 'x-some-header': 1 } },
+      { subject: 'Test' },
+      { subject: 'Test2' }
+    ]
+
+    it('should filter by differect predicates', () => {
```
pdobrosi, line 25: typo: different

pdobrosi (approved):
I also wanted to implement the same feature :) , I like your solution.

soulteary:
This feature looks very useful, I agree with @pdobrosi .

But the current code seems to have some conflicts, I will perform functional verification and resolve conflicts in a private fork.

If everything works, your name will appear in the v2.0.2 version, as a contributor, please continue to submit and play together @madflow 

@kowalczyka webui need update, we can talk it at https://github.com/maildev/maildev/issues/381

PR #364, opened 2021-06-27
Fix docker port mapping

agbaraka:
Docker container exposes port 25 and port 80 to be used for SMTP and Web client respectively and not port 1025 and 1080.

Therefore I have updated the command snippet to bind host ports 1025 and 8080 to container port 25 and 80 respectively.

Note: I used port 8080 instead of 1080 because binding to port 1080 in some systems will require administrative privilege


soulteary:
@agbaraka  @jcoutch Thanks for the feedback, the project has fixed this problem https://github.com/maildev/maildev/issues/382 , welcome to join us to continue to improve this project. https://github.com/maildev/maildev/issues/379

PR #368, opened 2021-10-19
fix: Program unknown argument

KhalitovAdel:


soulteary:
Thanks for the feedback, this feature is very useful. @KhalitovAdel 

The test seems to have no problem, but in fact, during the running process, the program cannot be automatically recycled, and there will be **a zombie process**.

It doesn't matter though, I'll merge this pr and fix it later.

This feature will be released in v2.0.2.

Issue #233, opened 2018-09-28
any plan to support bcc?


kimhanjoon:
Is there any plan to support bcc?

kgeis:
What do you mean by "support"?.

kimhanjoon:
When I send a mail to maildev server with bcc, maildev doesn't have bcc information. (or am I talking wrong?)
If maildev had bcc information, you would say "maildev supports bcc".

kgeis:
Thank you for the clarification. To explain my question, I could have imagined you want to bcc some address when relaying email.

refractalize:
would be good to see the SMTP "envelope" in the "Display" menu, i.e. the `MAIL FROM`, `RCPT TO` instructions (including BCC addresses) used when sending the email.

aandis:
+1

soulteary:
A community user submitted a functional implementation of bcc https://github.com/maildev/maildev/pull/369 , we will release it together in v2.0.2 version.

However, there are currently some problems with the interface style. I don't think it is a big problem. We can improve it in the next version.

Hope this news is helpful to you @aandis @kimhanjoon @kgeis @refractalize

PR #369, opened 2021-11-05
BCC support

delbertooo:
As maildev is primarily a debugging tool we want to validate the correctness of TO, CC and BCC as well.

See #233 

## Possible solutions
- multiply the emails for each recipient
- show cc and bcc addresses in the preview

Duplicating emails won't help debugging and may make it even worse. So we went for the "show in preview" option. 

## Implementation

Bcc should not be contained in the mail source. Otherwise it wouldn't be a "blind" copy as the recipients could see the blind copy recipients. Every recipient (to, cc, bcc) should be part of the `RCPT` though. So from my understanding and the insight of this [Mailtrap article](https://mailtrap.io/blog/cc-bcc-in-smtp/) bcc should be `allRecipients - to - cc`. This calculated bcc is what I implemented and added to the stored parsed email.

On the GUI side I found that it's actually hard to validate the recipients in maildev. Besides the headers this information isn't really shown in detail. To fix the missing details I implemented a panel below the email toolbar to show information. I'm not quite sure if this panel meets your UI expectations but in my opinion something like that would be a great addition to the project. This way the user can easily debug who should've received the mail.

image
image

## Preview

I built a docker image as a preview with the recipients/bcc support added by this pull requests. You can find it at [delbertooo/maildev-bcc](https://registry.hub.docker.com/r/delbertooo/maildev-bcc).

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -168,6 +168,7 @@
 
         <script src="scripts/app.js"></script>
         <script src="scripts/services.js"></script>
+        <script src="scripts/components/address.js"></script>
         <script src="scripts/controllers/main.js"></script>
         <script src="scripts/controllers/item.js"></script>
 
--- a/app/scripts/components/address.js
+++ b/app/scripts/components/address.js
@@ -1,0 +1,15 @@
+/* global app */
+
+/**
+ * This compoonent displays an address either with or without a name.
+ */
+
+app.directive('appAddress', function () {
+  return {
+    restrict: 'E',
+    scope: {
+      addr: '=address'
+    },
+    templateUrl: 'views/address.html'
+  }
+})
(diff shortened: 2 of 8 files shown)


soulteary:
Thanks for the feedback @delbertooo  I tested this function locally and found no problem. we will release it together in v2.0.2 version

However, there are currently some problems with the interface style. I don't think it is a big problem. We can improve it in the next version.

soulteary:
Thanks for the feedback @delbertooo  I tested this function locally and found no problem. we will release it together in v2.0.2 version

However, there are currently some problems with the interface style. I don't think it is a big problem. We can improve it in the next version.

delbertooo:
> However, there are currently some problems with the interface style. I don't think it is a big problem. We can improve it in the next version.

I didn't commit a rebuilt `app/styles/style.css` with `npm run css`. That's why the new styles are missing. I can do another merge request with rebuilt styles, if you want.

I guess I wasn't sure if that is the right way, as I'm not the biggest fan of binaries / artifacts in the repository.

PR #372, opened 2021-12-13
Add support for IMAP, POP3 and SMTP with Authentication

### LINKED ISSUE #166: Using another mail client of my choosing

jhf:
This PR adds dovecot with configuration for IMAP, POP3 and SMTP with authentication.

The Dovecot setup is pretty straight forward, without any ssl setup.
The added SMTP with authentication was to test our system with credentials for SMTP.

I did move the port from 1025 and 1080 to 25 and 80, as there is no longer a problem to bind to privileged ports
in docker, but I can revert that.

I use `dumb-init` to run both dovecot and `maildev`.

I've added some minimal documentation, and a CLI example of how to do SMTP with authentication.

This is inspired by the PR #209 , but is hopefully less intrusive by not moving around any js files.

This also fixes #166 .

I've published the packages - for testing - at `veridit/maildev` as we are using that for our own project.

To run locally and test one can try with `docker run --rm -it -p 1587:587 -p 1080:80 -p 1025:25 -p 1143:143 veridit/maildev`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Base
 FROM node:10-alpine as base
-MAINTAINER "Dan Farrelly <daniel.j.farrelly@gmail.com>"
+LABEL org.opencontainers.image.authors="Dan Farrelly <daniel.j.farrelly@gmail.com>"
 
 ENV NODE_ENV production
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,16 +18,39 @@ RUN npm install \
 # Prod
 FROM base as prod
 
-USER node
-WORKDIR /home/node
+RUN apk update \
+  && apk add dumb-init \
+  && apk add dovecot dovecot-submissiond \
+  && apk add sudo
+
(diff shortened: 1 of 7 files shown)


jhf:
After testing this solution, it seems that Maildev does not add files in the Maildir format 🤦‍♂️.
I assumed it did, due to the detailed description in PR #209.
Given that, we are not going to use Maildev, but rather setup a small docker image with SMTP and IMAP, based on the work from this PR.
I'll abandon this PR for Maildev.

PR #371, opened 2021-11-25
doc: corrected ports in readme

benobab:

PR #376, opened 2022-02-08
Fix docker port binding

campfireman:
Hey, thanks for creating maildev! I use it a lot for local development.

I encountered an issue with the docker command in the README and the docs. The server runs on port `25` and `80` within the container, but the README suggests binding port `1025` and `1080`. Therefore, the command doesn't work out of the box.

I hope the fix provided is valid for all systems and users, I tested it on Linux (5.10.89-1-MANJARO) with docker 20.10.12 (build e91ed5707e)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ If you want to use MailDev with [Docker](https://www.docker.com/), you can use t
 For a guide for usage with Docker,
 [checkout the docs](https://github.com/maildev/maildev/blob/master/docs/docker.md).
 
-    $ docker run -p 1080:1080 -p 1025:1025 maildev/maildev
+    $ docker run -p 1080:80 -p 1025:25 maildev/maildev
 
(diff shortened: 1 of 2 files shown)


campfireman:
Duplicate of #371

soulteary:
In fact, there is no problem with the current code of this repo. The problem is that the docker image uses the old code from two years ago.

I tried to create a new docker image based on the current code, and after running it, the default port is indeed as shown in the documentation. **So we don't have to modify the port number in the documentation.**

You can use this command to verify.

```bash
docker run -p 1080:1080 -p 1025:1025 soulteary/maildev:1.1.0
```

soulteary:
@campfireman Thanks for the feedback, the project has fixed this problem https://github.com/maildev/maildev/issues/382 , welcome to join us to continue to improve this project. https://github.com/maildev/maildev/issues/379

PR #383, opened 2022-03-23
feat: upgrade docker image and package.json

soulteary:
detail:

- fix dev script, update maintainer list, update version, update deps to fix npm fetch nodemailer
- update dockerfile
- release new docker image


soulteary:
@idears-org I think you are right, I will rewrite the image for node 16 to solve this hidden danger.
Maybe we can ignore npm warnings in node 16 (lockfile version upgrade)

PR #408, opened 2022-04-14
Use strip-js to sanitise html content

### LINKED ISSUE #398: Email display error
Test mail not show correct:

Console log error:

Same email show perfect with old maildev docker

Original test message attached:
[il3Q4M3X.eml.txt](https://github.com/maildev/maildev/files/8352547/il3Q4M3X.eml.txt)

dcrouch79:
Use [strip-js](https://www.npmjs.com/package/strip-js) to sanitise html content instead of `turndown` + `marked` libraries.

`strip-js` removes all javascript from the html while preserving everything else. The combination of `turndown` + `marked` to convert html to markdown and back corrupts the html in certain cases.

Other changes:
- Added a test to confirm that html table elements are not corrupted (fails in verison 2.0.2)
- Added another example email to send.js script containing a heavily styled html table
- Fixed other failing tests caused by using strip-js (better preservation of html content)

Fixes #398

### DIFF
--- a/lib/mailserver.js
+++ b/lib/mailserver.js
@@ -16,7 +16,8 @@ const logger = require('./logger')
 const smtpHelpers = require('./helpers/smtp')
 const { calculateBcc } = require('./helpers/bcc')
 const outgoing = require('./outgoing')
-const stripJs = require('strip-js')
+const createDOMPurify = require('dompurify')
+const { JSDOM } = require('jsdom')
 
 const store = []
 const eventEmitter = new events.EventEmitter()
@@ -267,7 +268,14 @@ mailServer.getEmail = function (id, done) {
 
   if (email) {
     if (email.html) {
-      email.html = stripJs(email.html)
+      // sanitize html
+      const window = new JSDOM('').window
+      const DOMPurify = createDOMPurify(window)
+      email.html = DOMPurify.sanitize(email.html, {
+        WHOLE_DOCUMENT: true, // preserve html,head,body elements
+        SANITIZE_DOM: false, // ignore DOM cloberring to preserve form id/name attributes
+        ADD_TAGS: ['link'] // allow link element to preserve external style sheets
+      })
     }
     done(null, email)
   } else {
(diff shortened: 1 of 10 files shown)


DominikSerafin:
[strip-js](https://github.com/shivanshu3/strip-js) is a no longer maintained package. If this was to be implemented then better solution would be to use [DOMPurify](https://github.com/cure53/DOMPurify).

DominikSerafin:
Also, `action` attribute shouldn't be removed from `form` tags - some email clients allow forms.

https://www.litmus.com/blog/the-ultimate-guide-to-interactive-forms-in-email/
https://css-tricks.com/html-forms-in-html-emails/

dcrouch79:
Ooops, didn't realise strip-js was deprecated (I should have checked github not just npmjs).

I actually tried DOMPurify initially but I struggled to get some of the existing tests to pass. I'll take another look now I'm a bit more familiar. I'll add another test and example that uses a form with action attribute too.

dcrouch79:
To get DOMPurify to work satisfactorily I had to enable some non-default config options:

```js
email.html = DOMPurify.sanitize(email.html, {
  WHOLE_DOCUMENT: true, // preserve html, head, style elements for inline styles
  SANITIZE_DOM: false, // ignore DOM cloberring to preserve form id/name attributes
  ADD_TAGS: ['link'] // allow link element to preserve external style sheets
})
```

Faq:
@soulteary any chance to get this reviewed?

djfarrelly:
Thank you @dcrouch79 - this looks fantastic! Love the DOMPurify package as well. We appreciate your work!

lib/mailserver.js
```diff
@@ -16,8 +16,8 @@ const logger = require('./logger')
 const smtpHelpers = require('./helpers/smtp')
 const { calculateBcc } = require('./helpers/bcc')
 const outgoing = require('./outgoing')
-const TurndownService = require('turndown')
-const marked = require('marked')
+const createDOMPurify = require('dompurify')
+const { JSDOM } = require('jsdom')
```
djfarrelly, line 20: Post-merge and npm publish, I realized that `jsdom` is not explicitly included in the package.json dependencies so this fails on the latest version on npm. This can work locally as jsdom is included as a dev dep's own dependency, but this isn't included in prod.

Issue #411, opened 2022-04-24
HTML page is not rendered properly


GaneshanNT:
The email template is not rendered properly in maildev inbox but the same email template is rendered properly if I ran directly in the browser.

Attached the screenshots for your reference

Maildev screen

<img width="992" alt="not rendered properly1" src="https://user-images.githubusercontent.com/26402630/164966488-71e41a56-fe24-4183-abf4-e0cc9633fd5d.png">

Email template

<img width="926" alt="rendered properly" src="https://user-images.githubusercontent.com/26402630/164966515-5e10f5d4-06f9-4352-9f9a-903d9a451aa8.png">

dablurr:
Hey guys, 

Having the same issues with the HTML on my end too. Using the `latest` image from Docker.

jkolarovic:
Hi everyone,
Same issue here !  :-)

Huluti:
Hello! Having the same issue :)

paxyknox:
We notice this issue happens when using swift and smtp but not gmail protocol.

djfarrelly:
@GaneshanNT or anyone else who has experienced the same issue - can you provide steps/code or an email body to reproduce the error? With that we could add it to the test script and fix this one up. Thanks!

danghieu90:
same issue https://github.com/maildev/maildev/issues/398

djfarrelly:
Fixed in https://github.com/maildev/maildev/releases/tag/v2.0.4

Couapy:
Tested, and it works well !
Thank's a lot for your work :+1:

GaneshanNT:
Thank you!! ... it is working fine now

pedrovagner:
When `latest` Docker image release will be updated?

Issue #403, opened 2022-04-02
All html attributes seem to be missing in HTML view


garth:
Using v2.0.2 docker image. In html view all html element attributes are stripped. When viewing the email src the attributes are present but encoded with "quoted-printable", maybe that is why they get stripped? v1.1.1 works fine.

beholderrk:
![image](https://user-images.githubusercontent.com/841755/161730414-df4c3709-5e0a-4b92-a874-732c232e65e1.png)
Yep, html email broken in 2.0.2 version

kossa:
Getting the same issue(without docker)

<img width="1091" alt="image" src="https://user-images.githubusercontent.com/1175584/162612399-e55479c5-286c-4d19-8ebc-3c958b37c9ad.png">

![image](https://user-images.githubusercontent.com/1175584/162612422-424c5519-0355-4887-96b8-1bbb00474328.png)

Mail preview:

![image](https://user-images.githubusercontent.com/1175584/162612432-e741aef7-e1f8-43c8-88fa-3b3fb941e505.png)

dcrouch79:
The `=3D` characters are how the [Quoted Printable](https://en.wikipedia.org/wiki/Quoted-printable) encoding system encodes the equals sign. See [here](https://stackoverflow.com/questions/4016067/whats-a-3d-doing-in-this-html) for more details.

The missing html elements/attributes is due to how the html is being sanitised. I've proposed a fix for this. See #398 and #408.

djfarrelly:
Fixed in https://github.com/maildev/maildev/releases/tag/v2.0.4

beholderrk:
@djfarrelly Thx!

PR #412, opened 2022-05-17
Ensure iframe is not undefined.

### LINKED ISSUE #398: Email display error

djfarrelly:
### Description

Prior to the 2.0 release, the `iframe` was set at the highest level of the closure, but it no longer is, so instead of setting and reading that variable, this passes it as an argument or searches the page for it.

Fixes #398, #411, #403

### DIFF
--- a/app/scripts/controllers/item.js
+++ b/app/scripts/controllers/item.js
@@ -4,172 +4,211 @@
  * Email item Controller -- The UI for the email pane
  */
 
-app.controller('ItemCtrl', [
-  '$scope', '$rootScope', '$routeParams', '$location', 'Email', '$http', '$cookies',
-  function ($scope, $rootScope, $routeParams, $location, Email, $http, $cookies) {
-    const iframe = null
-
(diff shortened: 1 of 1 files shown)


app/scripts/controllers/item.js
```diff
@@ -4,172 +4,211 @@
  * Email item Controller -- The UI for the email pane
  */
 
-app.controller('ItemCtrl', [
-  '$scope', '$rootScope', '$routeParams', '$location', 'Email', '$http', '$cookies',
-  function ($scope, $rootScope, $routeParams, $location, Email, $http, $cookies) {
-    const iframe = null
```
djfarrelly, line 10: Lots of auto-formatting in this file - this is what was removed and we no longer use this higher level variable in the closure.

PR #414, opened 2022-05-18
Fix CSS to display the last few lines of long emails correctly

clicube:
When the mail was long enough, the last few lines of the mail was not displayed.
I fixed CSS to fix this issue.

### Main cause
`.email-content` overflowed from parent container `.email-container`.

### Changes
- Set parent container `.email-container` to use flex box.
- Set `.email-content` to be resized adaptively.


clicube:
I think this will also fix #417.

DBLaci:
I confirm the iframe overflow with long mail content is fixed with this PR!

Until it is merged I use the docker image built from the fork - docker.io/dblaci/maildev-20221021 it only contains this pull commit and the mailder 2.0.5 as base. - maybe it helps someone temporarily

Weqqen:
Got the same problem. Hoping for merge soon :)

micter59:
Yes, same problem here with a mail which is too long. I'll test the image mentionned before.

leonardomnds:
Please, merge this.

clicube:
@djfarrelly Hi, part of the email is hidden, which continues to cause inconvenience. Is further work required to merge this PR?

assets/styles/style.scss
```diff
@@ -615,12 +615,16 @@ body.dark-theme {
...
+    height: 100vh;
+    display: flex;
+    flex-direction: column;
 }
 
 .email-content {
     box-sizing: border-box;
     z-index: 1;
-    height: calc(100vh - #{$toolbarHeight});
+    height: 0;
```
dimbslmh, line 626: `height: 0;` is unnecessary.

dimbslmh:
This will also fix https://github.com/maildev/maildev/issues/451

assets/styles/style.scss
```diff
@@ -615,12 +615,16 @@ body.dark-theme {
 
 
 .email-container {
+    height: 100vh;
```
jerome-rdlv, line 618: This should better be `height: 100%` because we really want this element to be 100% of the height of its parent. That would account for possible future changes in global layout.

assets/styles/style.scss
jerome-rdlv, line 626: For clarity, `height` should be removed completely, instead of `height: 0` which is overridden by `flex-grow: 1`.

jerome-rdlv (changes requested):
Hello. Two little amends in the PR for clarity and future proofness.

clicube:
Apologies for the delayed update. Thank you for the comment. I've accepted the feedback and added a new commit.

zemkogabor:
Are there plans to merge this pull request?

Freekbron:
A merge would be nice

alamirault:
How can we help to make this PR merged ?

Thanks !

PR #419, opened 2022-06-02
Fixed healthcheck with MAILDEV_BASE_PATHNAME

blur:
Added MAILDEV_BASE_PATHNAME env variable to healthcheck

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,4 +21,4 @@ ENV MAILDEV_WEB_PORT 1080
 ENV MAILDEV_SMTP_PORT 1025
 ENTRYPOINT ["bin/maildev"]
 HEALTHCHECK --interval=10s --timeout=1s \
-  CMD wget -O - http://localhost:1080/healthz || exit 1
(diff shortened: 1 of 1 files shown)


arjanfrans:
This is currently blocking me from using maildev, can this please get merged?

Dockerfile
```diff
@@ -21,4 +21,4 @@ ENV MAILDEV_WEB_PORT 1080
 ENV MAILDEV_SMTP_PORT 1025
 ENTRYPOINT ["bin/maildev"]
 HEALTHCHECK --interval=10s --timeout=1s \
-  CMD wget -O - http://localhost:1080/healthz || exit 1
\ No newline at end of file
+  CMD wget -O - http://localhost:1080${MAILDEV_BASE_PATHNAME}/healthz || exit 1
```
arjanfrans, line 24: Should the port also be a variable?

arjanfrans: I also do not think that this works with environment variables, see: https://stackoverflow.com/questions/67286967/dockerfile-healthcheck-with-environment-variable

PR #435, opened 2022-10-30
Add TLS support for incoming mails

### LINKED ISSUE #182: Support for TLS (via ngrok or somehow?)
On some services to allow custom smtp server, TLS is required. auth0.com is one example.

It seems that TLS is not supported in MailDev, so I thought of trying ngrok (pro account) to use tls

```
# start docker image
$ docker run --rm -p 1080:1080 -p 1025:1025 djfarrelly/maildev bin/maildev --web 1080 --smtp 1025 --incoming-user smtp --incoming-pass smtp

# start ngrok tunnel
$ ngrok tls -subdomain=SUBDOMAIN 1025
Forwarding                    tls://SUBDOMAIN.ngrok.io -> localhost:1025 

# manual check for connection
$ /usr/local/opt/openssl/bin/openssl s_client -starttls smtp -crlf -connect SUBDOMAIN.ngrok.io:443
CONNECTED(00000003)
didn't found starttls in server response, try anyway...
140735930754056:error:140790E5:SSL routines:ssl23_write:ssl handshake failure:s23_lib.c:177:
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 0 bytes and written 343 bytes
---
New, (NONE), Cipher is (NONE)
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : 0000
    Session-ID: 
    Session-ID-ctx: 
    Master-Key: 
    Key-Arg   : None
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    Start Time: 1496808447
    Timeout   : 300 (sec)
    Verify return code: 0 (ok)
---
```

So far I haven't been able to succeed. Is there a way you would suggest to add TLS over MailDev? It would be great for this development environment (ie: external service that really want TLS).

Thanks!

### LINKED ISSUE #429: tls for incoming mails
Hello,

Are there any plans to support TLS for incoming emails? It would be nice to be able to more easily make the transition from dev to production when sending emails.

sharat87:
Fixes #182
Fixes #429

This PR adds support for TLS for incoming SMTP connections. We add the following three new arguments to configure support for this. Unless the `--incoming-secure` option is set, _absolutely no_ change should be there from status quo.

| Options                          | Environment variable       | Description                                                                               |
| -------------------------------- | -------------------------- | ----------------------------------------------------------------------------------------- |
| `--incoming-secure`              | `MAILDEV_INCOMING_SECURE`  | Use SMTP SSL for incoming emails                                                          |
| `--incoming-cert <path>`         | `MAILDEV_INCOMING_CERT`    | Cert file location for incoming SSL                                                       |
| `--incoming-key <path>`          | `MAILDEV_INCOMING_KEY`     | Key file location for incoming SSL                                                        |

The contribution guidelines speak of checking lint result, which I've just done. Let me know if you need anything else from me to make this PR more aligned.

Thank you for all the tremendous work on maildev! 🙏


san-ouadghiri:
@djfarrelly, @soulteary can you consider this PR 👀? If it's good stuff, than several people would be happy as it would solve the TLS support topic. If not, any feedback to @sharat87 may unlock this.

lib/mailserver.js
```diff
@@ -170,19 +170,25 @@ mailServer.create = function (
...
     {
+      secure: isSecure,
+      cert: certFilePath ? fs.readFileSync(certFilePath) : null,
+      key: keyFilePath ? fs.readFileSync(keyFilePath) : null,
       onAuth: smtpHelpers.createOnAuthCallback(user, password),
       onData: handleDataStream,
       logger: false,
       hideSTARTTLS: true,
-      disabledCommands: user && password ? ['STARTTLS'] : ['AUTH']
+      disabledCommands: user && password ? [] : ['AUTH']
```
soulteary, line 191: It looks like a default behavior has been modified here, and there may be some compatibility issues. 

Would it be better to handle a separate config configuration based on the secure status? Perhaps a simple discussion is needed here?

sharat87: Hey, thanks for checking this out. Good point. I think something like this would be more backwards compatible:

```javascript
disabledCommands: user && password ? (isSecure ? [] : ['STARTTLS']) : ['AUTH']
```

I can push this change if it looks okay.

lib/mailserver.js
```diff
@@ -170,19 +170,25 @@ mailServer.create = function (
...
 ) {
   mailServer.mailDir = mailDir || defaultMailDir
   createMailDir()
 
   const hideExtensionOptions = getHideExtensionOptions(hideExtensions)
   const smtpServerConfig = Object.assign(
     {
+      secure: isSecure,
+      cert: certFilePath ? fs.readFileSync(certFilePath) : null,
+      key: keyFilePath ? fs.readFileSync(keyFilePath) : null,
```
jan--f, line 186: isn't maildev currently using self signed certs for this? Can we keep this behavior as a fallback?

jan--f:
Thanks, I really like this PR. This would also fix #449 and consequently superseed #450 

Fixes: #449

lib/mailserver.js
sharat87: I don't think it does. From what I can tell from my understanding of the code, here's any support for SSL, self-signed or not, currently. So this shouldn't be breaking any existing configurations.

lib/mailserver.js
djfarrelly: This looks good to me - I'll just add this change and get this merged!
 
```suggestion
      disabledCommands: user && password ? (isSecure ? [] : ['STARTTLS']) : ['AUTH']
```

djfarrelly (approved):
This looks great - I'll get the additional change that @soulteary and @sharat87 suggested on the main branch

PR #441, opened 2022-12-27
feat: added arm v8 to docker build and push

JamesMarino:
Added ARM v8 Docker architecture to both build and push scripts.

This has been tested using a personal repo.

### DIFF
--- a/scripts/dockerBuild.sh
+++ b/scripts/dockerBuild.sh
@@ -3,7 +3,7 @@
 #   jq (https://stedolan.github.io/jq/)
 
 # Build cross platform by default
-DEFAULT_PLATFORM="linux/amd64,linux/arm64"
+DEFAULT_PLATFORM="linux/amd64,linux/arm64,linux/arm/v8"
 PLATFORM="${1:-$DEFAULT_PLATFORM}"
 
 VERSION=`npm version --json | jq -r .maildev`
--- a/scripts/dockerPush.sh
+++ b/scripts/dockerPush.sh
@@ -3,7 +3,7 @@
 #   jq (https://stedolan.github.io/jq/)
 
 # Build cross platform by default
-DEFAULT_PLATFORM="linux/amd64,linux/arm64"
+DEFAULT_PLATFORM="linux/amd64,linux/arm64,linux/arm/v8"
 PLATFORM="${1:-$DEFAULT_PLATFORM}"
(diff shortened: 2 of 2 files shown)


djfarrelly (changes requested):
Hi @JamesMarino - I'd love to support this change, but my latest Docker for Mac doesn't have `linux/arm/v8` support and I'd like to test this myself before merging (so I can also publish the images). Do you have any information on how one needs to install the v8 support to build these images? I spent some time trying to sort this out and I couldn't find anything that worked for me - Also I'd like to include that info within the `dockerBuild.sh` file for future devs that need to run that script!

JamesMarino:
Thanks for the review @djfarrelly - what error were you getting when trying to run the build locally? I'm not sure there is anything I needed to do in particular to support `linux/arm/v8` on my Mac. I'm able to run the command on Docker Desktop `4.15.0`, Engine `20.10.21` and BuildX `0.9.1` and on a MacBook Pro with Intel and Apple Silicon chips.

Addressing the issue higher, how would you feel about running a GitHub Action that would automatically Docker build and push to the `maildev/maildev` Docker Hub when run on master? This would take out the problem for issues arising when building and pushing multiple platforms on individual devices locally.

djfarrelly:
Hi @JamesMarino, my Docker for mac gives me these options for buildx:

```
docker run --privileged --rm tonistiigi/binfmt --install all
{
  "supported": [
    "linux/arm64",
    "linux/amd64",
    "linux/riscv64",
    "linux/ppc64le",
    "linux/s390x",
    "linux/386",
    "linux/mips64le",
    "linux/mips64",
    "linux/arm/v7",
    "linux/arm/v6"
  ],
  "emulators": [
    "qemu-arm",
    "qemu-i386",
    "qemu-mips64",
    "qemu-mips64el",
    "qemu-ppc64le",
    "qemu-riscv64",
    "qemu-s390x",
    "qemu-x86_64"
  ]
}
```

I would be up for automating pushing to Docker hub, the only challenges are with QA on the image (less needed if there are automated integration tests with the container image. I think the image release can be run on Github actions, but should still be triggered for a particular release tag then just push to "latest" (I've also seen that having version tags on Docker images are best if a regression sneaks into `master` it might break people's setup). Would you be interested in writing that Github action for this repo?

PR #444, opened 2023-01-19
Document the existing `removeListener` method

illright:
This method is exposed on the `maildev` instance, but not documented, even though it is useful to avoid duplicate subcriptions


djfarrelly:
Thanks! - This will be covered in #497

Issue #449, opened 2023-02-23
Option to have STARTTLS


jan--f:
In #274 its was argued that STARTTLS is completely broken and should not be advertised. [A comment](https://github.com/maildev/maildev/issues/274#issuecomment-1193144387) on that issue mentioned that this works fine for the Prometheus community. Specifically it works when username and password are _not_ set.

I'd like to raise this again, since for prometheus/alertmanager we still use maildev v1.1.0 and we'd love to use something newer.
Before #276 it was already possible to configure maildev to hide STARTTLS, so I'm not sure why it was completely disabled.
Disabling it and adding an option to advertise STARTTLS would also be an option.

jan--f:
Based on the https://github.com/maildev/maildev/issues/274#issuecomment-1193144387 I opened a PR. With this alertmanager tests pass against an up to date maildev again. Specifically [this test](https://github.com/prometheus/alertmanager/blob/6450c9ff18a4f7a6ebcefa83e157e80485014a4b/notify/email/email_test.go#L384-L415).

djfarrelly:
Closed with #435.

PR #450, opened 2023-02-23
mailserver: only hide STARTTLS when user+password is configured

jan--f:
Fixes: #449


djfarrelly:
Superseded by #435 - thanks for your contribution @jan--f!

PR #469, opened 2023-08-29
Add option to enabe/disable hideSTARTTLS

### LINKED ISSUE #468: maildev does not allow STARTTLS advertising
The smpt_server is create with hardcoded [`hideSTARTLS: true`](https://github.com/maildev/maildev/blob/357a20edcd205413d3590aedb8fcd7c97563c40d/lib/mailserver.js#L190). It would be better to expose this via a config option for tests where a client specifically check whether starttls is advertised.

jan--f:
Fixes: #468

### DIFF
--- a/index.js
+++ b/index.js
@@ -38,6 +38,7 @@ module.exports = function (config) {
     config.incomingUser,
     config.incomingPass,
     config.hideExtensions,
+    config.hideStarttls
     config.incomingSecure,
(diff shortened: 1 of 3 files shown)


jan--f:
This is an ammendment to #435. @djfarrelly @soulteary wdyt? This keeps the current default behavior but adds an option to enable STARTTLS advertising.

PR #471, opened 2023-09-27
spring boot configuration added

dipu11:
To use the SMTP server for the spring boot app, I have added the configuration for the `application.properties` file or `application.yml` file

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,38 @@ config.action_mailer.delivery_method = :smtp
 
 **Drupal** -- Install and configure [SMTP](https://www.drupal.org/project/smtp) module or use a library like [SwiftMailer](http://swiftmailer.org/).
 
+**Spring Boot** -- configuration:
+<br/>
+in application.properties file:
+```
+spring.mail.host=localhost #where the smtp server is running
+spring.mail.port=1025
+spring.mail.username=no-reply@gmail.com
+spring.mail.properties.mail.smtp.starttls.enable=true
+spring.mail.properties.mail.smtp.starttls.required=true
+spring.mail.properties.mail.smtp.auth=true
+spring.mail.properties.mail.smtp.connectiontimeout=5000
+spring.mail.properties.mail.smtp.timeout=5000
+spring.mail.properties.mail.smtp.writetimeout=5000
+```
+Or in application.yml file:<br/>
+```
+spring:
+    mail:
+        properties:
+            mail:
+                smtp:
+                    starttls:
+                        enable: true
+                        required: true
+                    auth: true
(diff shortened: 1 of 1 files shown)


dancer1325:
Hi @dipu11 , do you have some demo project to check that it works? Because I would expect some error throw by the `javax.mail` or whatever SMTP interface you use, if you specify `starttls`, without passing `spring.mail.username` and `spring.mail.password`

dipu11:
@dancer1325 , I have used this same configuration(without setting user/pass) in at least two production-grade projects for testing emails.
Unfortunately, I don't have such demo projects to link here.

Frequency21:
@dipu11 Would you mind to create a demo project or take a glance at my configs or share some tips, please?

I am trying to setup a local spring (5.3.4) application to send emails to a local smtp server (maildev) but it fails to connect when I set MAILDEV_INCOMING_SECURE. (Without TLS related options it works perfectly)

Also I use mkcert which installs the rootCA to system and java truststore. So I am pretty sure that the self-signed cert is valid. (I can open the maildev site with https on chrome)

My configurations:
```env
MAILDEV_SMTP_PORT=587

MAILDEV_OUTGOING_HOST=192.168.100.73
MAILDEV_OUTGOING_PORT=465

MAILDEV_HTTPS=true
MAILDEV_OUTGOING_SECURE=true
MAILDEV_INCOMING_SECURE=true
MAILDEV_INCOMING_CERT=.\certs\maildev.pem
MAILDEV_INCOMING_KEY=.\certs\maildev-key.pem
MAILDEV_HTTPS_CERT=.\certs\maildev.pem
MAILDEV_HTTPS_KEY=.\certs\maildev-key.pem
MAILDEV_INCOMING_USER=test
MAILDEV_INCOMING_PASS=test

MAILDEV_IP=192.168.100.73
MAILDEV_WEB_IP=192.168.100.73

NODE_EXTRA_CA_CERTS=...\mkcert\rootCA.pem
```

Spring:
```yaml
spring:
  mail:
    host: 192.168.100.73
    port: 587
    protocol: smtp
    properties:
      mail:
        smtp:
          auth: true
          starttls.enable: true
          starttls.required: true
          timeout: 5000
          writetimeout: 5000
          connectiontimeout: 5000
    username: test
    password: test
    test-connection: true
```

Frequency21:
I solved my issue by adding an extra configuration to spring. 

```yaml
spring:
  mail:
    properties:
      mail.smtp:
          socketFactory.class: javax.net.ssl.SSLSocketFactory
```

Not sure yet why it was needed.

PR #476, opened 2023-12-19
The bottom content of the email is hidden because of the meta bar

benavern:
There is a portion of the emails that we can't see because of the set height of `.email-content`

Here I just make a flex box of the entire column , set it to 100% of its container and say `.email-content` must grow inside the available space (all the space under the `.email-header` and `.email-meta`). then its content can still be scrolled but we can see the entire content of the email

### DIFF
--- a/assets/styles/style.scss
+++ b/assets/styles/style.scss
@@ -615,13 +615,15 @@ body.dark-theme {
 
 
 .email-container {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
 }
 
 .email-content {
     box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


MargaritaSav (approved):

rwaness (approved):

clicube:
Same as #414

benavern:
ah yes, didn't look at the issues, sorry, I was just bothered by our testers taht thought the bug was on our side on threir first look 😅

clicube:
@benavern thanks, we hope this issue to fix :D

PR #478, opened 2023-12-21
feat: possibility to set timezone

rizlas:
As the title says, this PR aims to add the possibility to set container timezone.


djfarrelly (approved):
Great idea!

PR #495, opened 2024-05-31
Fix auth with no password

mikhailevtikhov:
There is an issue for this problem https://github.com/maildev/maildev/issues/436

The problem is that when logging in to an smtp server with username and password configured, you can log in by passing an arbitrary username and an empty password. This is due to an error in the authorization logic implemented in the file smtp.js.

The playback path:

```docker run -p 1080:1080 -p 1025:1025 --name maildev maildev/maildev --incoming-user test --incoming-pass password```

```telnet localhost 1025

EHLO localhost

AUTH LOGIN
334 VXNlcm5hbWU6
admin
334 UGFzc3dvcmQ6

235 Authentication successful
```

This PR fixes this bug

### DIFF
--- a/lib/helpers/smtp.js
+++ b/lib/helpers/smtp.js
@@ -6,12 +6,15 @@ const smtpHelpers = module.exports = {}
  * Authorize callback for smtp server
  */
 smtpHelpers.createOnAuthCallback = function (username, password) {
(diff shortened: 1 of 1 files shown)


djfarrelly:
Thanks for the fix!

PR #500, opened 2024-09-04
Fix Docker healthcheck host

rbwsam:
- Docker healthcheck command was using "localhost" but that resolves to ::1 (IPv6) so healthchecks failed.
- Changed the healthcheck to use "127.0.0.1" and healthchecks work as intended.


Faq:
https://github.com/maildev/maildev/pull/475

PR #502, opened 2024-10-12
fixed Docker run command

tanvirakon:
I believe the '$' symbol is redundant here as terminal has already one. So after coping the command with copy button we had to manually remove one '$'. 

So I removed the '$'.


3XC1T3D:
the '$' symbol is a common hint - that you should use the command in a terminal ^^

PR #504, opened 2024-11-28
Pagination support for getting all emails

little-mystic-stuff:
We are using MailDev for automated testing purpose.
We have a lot of emails, and we should take a "special email" via REST API.
Unfortunately, existing methods can return only all emails or already-known email.
The problem can be solved like a "get all emails, validate it, repeat until match or timeout". But it's not an effective way.

I propose the way described below.
- get all emails, validate it, **remember the number of emails**
- get all emails excepts the remembered number of emails (counts from beginning of storage)

The new REST endpoint allows it. It allow to read emails by groups, don't read the same emails many times, save traffic and time.

### DIFF
--- a/docs/rest.md
+++ b/docs/rest.md
@@ -71,6 +71,8 @@ Returns:
 
 **GET    /email** - Get all emails
 
+**GET    /email/skip/:number** - Get all emails excepts {number} of emails according to lifetime
+
 **DELETE /email/all** - Delete all emails
 
 **GET    /email/:id** - Get a given email by id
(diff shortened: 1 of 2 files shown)


djfarrelly (commented):
I love the idea here, but I'd rather implement this using a more REST-ful query string param like `?skip=` 

I'll fork and adjust!

djfarrelly:
@little-mystic-stuff - what do you think about this change combining into a single endpoint? https://github.com/maildev/maildev/pull/505

little-mystic-stuff:
@djfarrelly Hello, sorry, I couldn't answer in time.
I glad you like this idea.
Yep, I wanted to use query-parameter, but I'm not a pro in express-framework :) So I just offered what I could do.
Thanks for merging and thanks for this tool at all!

PR #536, opened 2026-07-07
Fix attachment download links

### LINKED ISSUE #513: Filenames are displayed with hashed name in web interface
Hi @djfarrelly,

The problem with #508 is that the filenames displayed in the web interface are now also hashed, and testers can't easily validate that attachments have correct filenames.

I don't know if this is possible to save the file on the server with the hashed name but still present the file in the web interface with its original name.

ahaly-compass:
## Summary

Fixes attachment downloads from the web UI by using the generated attachment filename in the download URL.

## Details

Attachment files are stored and resolved by their generated filename, but the UI was building download URLs from the original display filename. This caused attachment download requests to return 404 when those values differed.

The UI still displays the original attachment filename; only the URL identifier changes.

Fixes #513.

## Validation

- Built @maildev/core first so workspace types are available
- Ran `pnpm --filter @maildev/ui typecheck`
- Ran `pnpm --filter @maildev/ui lint`

### DIFF
--- a/packages/ui/src/components/email-viewer/EmailHeader.tsx
+++ b/packages/ui/src/components/email-viewer/EmailHeader.tsx
@@ -324,7 +324,7 @@ export function EmailHeader({ email }: EmailHeaderProps) {
           {email.attachments.map((attachment, index) => (
             <Tooltip key={index} content={`Download ${attachment.filename ?? 'attachment'}`}>
               <a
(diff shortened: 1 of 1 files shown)


djfarrelly (approved):
Thanks for this fix @ahaly-compass! Merging and it'll be in the next release

PR #539, opened 2026-07-08
Add bulk email delete endpoint

### LINKED ISSUE #520: Add deleteEmails endpoint
Hey maildev team,

I'd be keen to see the addition of a new endpoint similar to deleteEmail but it can accept an array of IDs so we can delete multiple emails in one go rather than spamming deleteEmail.

Thanks!

IFtech-A:
## Summary
- Add POST /api/email/delete for deleting multiple emails by ID in one request
- Return deleted and notFound ID lists so clients can handle partial matches
- Reuse the existing SMTP delete path when available so delete events still emit
- Document the endpoint and add a patch changeset for @maildev/api

Fixes #520

## Verification
- corepack pnpm --filter @maildev/api test
- corepack pnpm --filter @maildev/api typecheck
- corepack pnpm --filter @maildev/api lint
- corepack pnpm --filter @maildev/api build
- End-to-end checked with the local dev API/SMTP stack using synthetic emails

### DIFF
--- a/.changeset/add-bulk-delete-emails.md
+++ b/.changeset/add-bulk-delete-emails.md
@@ -1,0 +1,5 @@
+---
+'@maildev/api': patch
+---
+
+Add a bulk delete endpoint for deleting multiple emails by ID.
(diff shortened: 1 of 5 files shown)


djfarrelly:
Thanks @IFtech-A - this is a great contribution!

PR #548, opened 2026-07-27
chore: Keep MailDev responsive with tens of thousands of emails

dreic:
## Problem

With a large inbox (10,000+ emails) MailDev becomes unresponsive, and the only
way to recover is to delete the mail directory and restart the service.

I reproduced it before changing anything. Three problems compound, measured at
10,000 emails with ~12 KB bodies:

1. **The web UI refetched every email in full, every 5 seconds.** `GET /api/email`
   returns all emails with `html`, `text` and `headers` included — a **117 MB**
   response taking ~1 s of pure serialisation, on a `refetchInterval: 5000`
   plus an invalidation per arriving message. At 30,000 emails the process is
   OOM-killed outright (exit 137) trying to build that response.
2. **`MemoryStorage` was array-backed.** `getById`, `save` and `delete` each
   scanned the array, so ingesting *n* emails was O(n²), and `markAllRead` was a
   `getAll()` followed by a `save()` per email (255 ms at 10k).
3. **Nothing ever deleted `.eml` files.** With the default in-memory store the
   SMTP server still writes every message to disk permanently — which is exactly
   why wiping the directory is the only recovery, and why restarts keep getting
   slower.

## Results

| At 10,000 emails | Before | After |
| --- | --- | --- |
| List request (what the UI polls) | 984 ms, **117 MB** | 17 ms, **21 KB** |
| `PATCH /api/email/read-all` | 255 ms | 0.8 ms |
| Save 20,000 emails | ~1070 ms | 41 ms |
| `getById` × 2000 | ~180 ms | 0.6 ms |
| At 30,000 emails | **process killed** | fine |

## Changes

**`@maildev/core`** — `MemoryStorage` is backed by an insertion-ordered `Map`, so
lookup, save and delete are O(1) and the oldest email is always the first entry,
which makes eviction cheap. The unread count is maintained incrementally instead
of scanned. New operations: `list`, `listSummaries`, `markAllRead`, `stats`,
`onEvicted`.

**Bounded storage** — a `maxEmails` limit (default **1000**, `--max-emails 0` to
disable) drops the oldest email *along with its `.eml` file and attachments*
through an awaited `onEvicted` hook, so neither memory nor the mail directory
grows without bound. Backlogs left by earlier runs are pruned at startup, and
restoring from disk loads only the newest `maxEmails` in bounded-concurrency
chunks. Filesystem fan-out is capped throughout to avoid `EMFILE` on large
directories.

**`@maildev/api`** — new `GET /api/email/summary` returns a bounded page
(default 50, max 200) of body-free summaries with server-side `search`, `sort`,
`unread` and the counts a list view needs. `GET /api/email` keeps its existing
semantics but now accepts `limit` and `sort`. `read-all` routes through
`storage.markAllRead()`. Socket.io listeners are registered once rather than per
connection (they previously accumulated per open tab and tripped the
max-listeners warning), and `newMail` now carries a summary rather than the full
email.

**`@maildev/ui`** — the list renders one page at a time with infinite scroll
instead of a row per email in the store; the 5-second full refetch is gone;
search is debounced and runs on the server; socket updates are coalesced so a
burst of mail causes one refetch rather than one per message.

**`@maildev/mcp`** — the `maildev://emails` resource pretty-printed every email
in full, and the stdio client fetched the whole inbox over HTTP. Both are now
capped and the resource returns summaries.

## ⚠️ Behaviour change

**The 1000-email default discards older mail, where previously nothing was
discarded.** When emails are persisted, their `.eml` files and attachments are
deleted along with them.

I went with a bounded default because an unbounded store is the failure being
fixed here — but it is a data-loss-shaped change for existing users, so it's
worth a maintainer decision. `--max-emails 0` restores the old behaviour, and
I'm happy to flip the default to `0` (making the cap opt-in) if you'd prefer
that for a release.

`Storage` implementations must now provide `list`, `listSummaries`,
`markAllRead`, `stats` and `onEvicted`. In-repo that's `MemoryStorage` and
`FileStorage`; third-party implementations would need updating.

## Testing

`pnpm build`, `pnpm lint`, `pnpm typecheck` and `pnpm test` all pass. **278 tests,
up from 204** — the 74 new ones cover:

- eviction, the awaited evict hook, unread-count bookkeeping, pagination,
  search, sorting and stats
- `FileStorage` deleting files on delete / deleteAll / eviction, and staying
  bounded across many deliveries
- `pruneMailDir`, and `loadMailsFromDirectory` restoring only the newest N
- the summary endpoint: page bounds, limit clamping, malformed params,
  `limit=0`, search, unread filter, and route precedence over `/email/:id`
- the paginated UI list: one page rendered, next page on scroll, no further
  requests once exhausted, server-side search
- scale guards at 20,000 emails whose budgets sit ~10× above current timings but
  well below the old ones, so a reintroduced per-operation scan fails them,
  plus payload-size assertions that don't depend on timing at all

Beyond unit tests I ran the built CLI end to end: 2000 emails delivered, store
and directory both held at exactly the limit (4.7 MB, not GBs), a restart with a
smaller limit pruned 175 stale files, no errors logged. Then drove the web UI in
a browser — 50 rows rendered out of 300, a single `?limit=50` request,
`skip=50` on scroll, search hitting the server, full bodies fetched only when an
email is opened, and a 200-email burst producing 4 coalesced refetches rather
than 200.

`scripts/bench.js` is included for reproducing the numbers locally:

```bash
pnpm build && node scripts/bench.js
```

A changeset is included.


djfarrelly:
Thanks for the contribution @dreic! I liked ideas in this changeset, but wanted to break this into smaller atomic changes and limit some API surface area for now. The equivalent changes have been added in #557, #558, and #559. I appreciate your contributions!

PR #561, opened 2026-08-18
Revert "Revert "perf(core): back MemoryStorage with an insertion-ordered Map (1/3)""

djfarrelly:
Reverts maildev/maildev#560


djfarrelly:
Not needed — #557 is now cleanly on main in the correct order (#557 → #558 → #559). Closing.

PR #569, opened 2026-08-26
feat: record relay delivery status on the email

### LINKED ISSUE #199: Know if mail has been relayed
Hi,

Is it possible now to know if mail has been relayed ? Could be a new feature ?

ChrisJr404:
Closes #199. After a successful relay this sets `relayedAt` and `relayedTo` on the stored email and persists it, so the REST API can report whether a message was relayed and to which recipients. It hooks into `relayEmail`, which is the shared path for both auto-relay and the manual relay endpoint, and adds a test.

### DIFF
--- a/packages/core/src/types/email.ts
+++ b/packages/core/src/types/email.ts
@@ -101,6 +101,10 @@ export interface Email {
   attachments: Attachment[]
   /** SMTP envelope data */
   envelope: Envelope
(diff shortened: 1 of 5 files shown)


djfarrelly (approved):
Thanks for the contribution @ChrisJr404 - I am going to stack some PRs on top of this adding this info to the UI and other aspects.

PR #570, opened 2026-08-28
feat: persist relay delivery status across restarts

djfarrelly:
Stacked on #569 (`relay-status`). **Review/merge #569 first.**

## What
`relayedAt`/`relayedTo` (added in #569) were held only in memory, so restarting MailDev lost the record of whether a message had been relayed and to whom. This persists that status for disk-backed storage — via a **generic per-email metadata sidecar**, not a relay-specific file, so it can hold future persisted fields without adding new file types.

## How
- **`@maildev/core` `FileStorage`** writes a `<id>.meta.json` sidecar next to the `.eml` whenever an email carries persisted metadata. The raw `.eml` can't hold this — it's post-receipt state that isn't recoverable by re-parsing the message.
- The store routes every mutation through `save(email)`, so the sidecar is written as a **wholesale projection** of the email's persisted fields (`buildMetadata`) — no partial-update/merge logic needed. Extending it later = add one field to `EmailMetadata` + one block in `buildMetadata`/`readMetadata`.
- A new exported `EmailMetadata` type describes the persisted shape (today `{ relayedAt, relayedTo }`).
- The sidecar is removed alongside the email on `delete`, `deleteAll`, and `maxEmails` eviction.
- New optional `Storage.readMetadata(id)` exposes the persisted metadata; in-memory storage does not implement it (nothing to restore).
- **`@maildev/smtp`** `loadMailsFromDirectory` reads the sidecar and repopulates the email when restoring on startup.

## Tests
- `@maildev/core`: sidecar write/read round-trip, no sidecar when there's nothing to persist, cleanup on delete + deleteAll + eviction.
- `@maildev/smtp`: end-to-end — relay via a `FileStorage`-backed server, then restore from the same directory with a fresh store and assert the status survives.

Changeset included (patch: core, smtp).

### DIFF
--- a/packages/core/src/storage/file.ts
+++ b/packages/core/src/storage/file.ts
@@ -1,7 +1,7 @@
-import { mkdir, rm, readdir } from 'node:fs/promises'
+import { mkdir, rm, readdir, readFile, writeFile } from 'node:fs/promises'
 import { join, resolve, sep } from 'node:path'
 import { tmpdir } from 'node:os'
-import type { Email, StorageOptions } from '../types/index.js'
+import type { Email, EmailMetadata, StorageOptions } from '../types/index.js'
 import { mapLimit } from '../utils/concurrency.js'
 import { MemoryStorage } from './memory.js'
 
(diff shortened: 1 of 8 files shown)


packages/core/src/storage/file.ts
```diff
@@ -127,6 +137,59 @@ export class FileStorage extends MemoryStorage {
...
+  /**
+   * Write the relay-status sidecar for an email
+   */
+  private async writeRelayMetadata(email: Email): Promise<void> {
+    const payload = JSON.stringify({
+      relayedAt: email.relayedAt,
+      relayedTo: email.relayedTo ?? [],
+    })
+    await writeFile(this.getRelayMetadataPath(email.id), payload, 'utf8')
+  }
```
djfarrelly: May want to make this more generic - not just relay, but more `<id>.metadata.json` for future proofing

PR #580, opened 2026-09-10
fix: reject start() when the SMTP port is already bound (#568)

### LINKED ISSUE #568: Programmatic API: SMTP port conflict crashes the process instead of rejecting start()
## Summary

If the SMTP port is unavailable, `await maildev.start()` never rejects. The error is re-thrown from inside an asynchronous `error` handler, so it escapes as an uncaught exception and takes the process down — a `try`/`catch` around `start()` does not catch it.

This is the same scenario as #298 (embedding MailDev, misconfigured/occupied port, process dies with nothing pointing at MailDev), reaching the same outcome through a different mechanism in the v3 code.

## Cause

`handleServerError` special-cases `ECONNRESET`, then re-emits and re-throws everything else:

https://github.com/maildev/maildev/blob/main/packages/smtp/src/server.ts#L812-L821

```ts
private handleServerError(err: NodeJS.ErrnoException): void {
  if (err.code === 'ECONNRESET' && err.syscall === 'read') {
    // ...ignored
  } else {
    this.emit('error', err)
    throw err
  }
}
```

It is invoked from the server's `error` event:

https://github.com/maildev/maildev/blob/main/packages/smtp/src/server.ts#L106-L108

Two things go wrong there:

- `this.emit('error', err)` on an `EventEmitter` with no `error` listener throws by design.
- `throw err` is inside an event callback, not inside the `start()` promise, so it cannot reject that promise.

Meanwhile `start()`'s own `reject` is only wired to the `listen` callback:

https://github.com/maildev/maildev/blob/main/packages/smtp/src/server.ts#L111-L122

`EADDRINUSE` arrives via the `error` event rather than that callback, so the promise stays pending forever while the process exits.

## Reproduction

```js
import net from 'node:net'
import { MailDev } from 'maildev'

const PORT = 2711
await new Promise((r) => net.createServer().listen(PORT, '::', r))

try {
  const maildev = new MailDev({ smtp: PORT, disableWeb: true, silent: true })
  await maildev.start()
  console.log('start() resolved')
} catch (e) {
  console.log('start() rejected with', e.code)   // never runs
}
```

**Expected:** `start()` rejects with the `EADDRINUSE` error, and the caller decides what to do.

**Actual:** the `catch` never runs; the process dies:

```
node:events:486
      throw er; // Unhandled 'error' event
      ^
Error: listen EADDRINUSE: address already in use :::2711
    at Server.setupListenHandle [as _listen2] (node:net:1948:16)
    at listenInCluster (node:net:2005:12)
Emitted 'error' event on SMTPServer instance at:
    at SMTPServer.handleServerError (.../packages/smtp/dist/server.js:651:18)
  code: 'EADDRINUSE', errno: -48, syscall: 'listen', address: '::', port: 2711
```

Note the stack points into `node:net` and `smtp-server`; nothing in it suggests "your MailDev port is taken", which is what made #298 hard to diagnose.

## Suggested fix

Route startup errors to the `start()` promise, and stop re-throwing from the event handler:

```ts
async start(): Promise<void> {
  const config = this.buildServerConfig()
  this.smtp = new SMTPServerLib(config)

  return new Promise((resolve, reject) => {
    const onStartupError = (err: Error) => reject(err)
    this.smtp!.once('error', onStartupError)

    this.smtp!.listen(this.port, this.host, (err?: Error) => {
      if (err) return reject(err)
      // Startup succeeded: hand later errors to the normal handler.
      this.smtp!.off('error', onStartupError)
      this.smtp!.on('error', (e) => this.handleServerError(e))
      resolve()
    })
  })
}
```

and in `handleServerError`, drop the `throw` and only `emit('error')` when there is a listener:

```ts
} else {
  this.logger.error(err.message)
  if (this.listenerCount('error') > 0) this.emit('error', err)
}
```

That keeps errors observable for callers who want them, without a bare `emit` on a listener-less emitter being fatal — which is also the ask in #298.

## Why it matters

The most likely place to hit this is a test suite's `globalSetup`. A stray `maildev` left running on 1025 kills the whole run with a stack trace that implicates neither MailDev nor the port, and adding a `try`/`catch` — the obvious first instinct — does not help.

## Environment

- `maildev` 3.0.0-rc.1 (built from `main`)
- Node.js v24.14.0
- macOS (Darwin 25.2.0)

cpruijsen:
## Summary

- If the SMTP bind fails (`EADDRINUSE`, `EACCES`, …), `SMTPServer.start()` now rejects with that error instead of hanging forever.
- `handleServerError` no longer re-throws. Post-listen errors (`ETIMEDOUT`, …) are logged, and `error` is emitted only when a caller has subscribed — so a `try`/`catch` around `maildev.start()` actually works.
- `port: 0` (ephemeral bind) is unchanged.

Closes #568.

## Decision

- **Chose:** attach `once('error')` for the duration of `listen()`, then swap to `handleServerError`; never throw from the handler; `emit('error')` only if there is a listener.
- **Alternative:** a `listening` flag and a stored `reject` inside a single `handleServerError`.
- **Why:** that is the shape described on #568. Same behaviour, slightly clearer lifecycle.
- Happy to switch to the single-handler version if you prefer it.

No changeset in this PR. Happy to add a patch bump for `@maildev/smtp` / `maildev` if you want this in the rc changelog.

## Test plan

- [x] Occupy a port, `new SMTPServer({ port, host: '127.0.0.1' })`, `await start()` rejects with `{ code: 'EADDRINUSE' }`; no `uncaughtException`.
- [x] After a successful listen, a fake `ETIMEDOUT` on the smtp-server instance does not throw; with an `error` listener it is forwarded.
- [x] Existing `port: 0` tests still pass.
- [x] `pnpm --filter @maildev/smtp test` — 10 files, 64 tests passed.
- [ ] Optional: `new MailDev({ smtp: occupied, disableWeb: true, silent: true })` — `await maildev.start()` rejects; process stays up.

### DIFF
--- a/packages/smtp/src/__tests__/start-errors.test.ts
+++ b/packages/smtp/src/__tests__/start-errors.test.ts
@@ -1,0 +1,109 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import { mkdtemp, rm } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import net, { type AddressInfo } from 'node:net'
+import { MemoryStorage } from '@maildev/core'
+import { SMTPServer } from '../server.js'
+
+/** Drive smtp-server's 'error' event, the same path handleServerError listens on. */
+function emitSmtpError(server: SMTPServer, err: NodeJS.ErrnoException): void {
+  const smtpLib = (
+    server as unknown as {
+      smtp: { emit: (event: 'error', err: NodeJS.ErrnoException) => void }
+    }
+  ).smtp
+  smtpLib.emit('error', err)
+}
+
+describe('SMTPServer start() errors', () => {
+  let mailDir: string
+  let storage: MemoryStorage
+  let server: SMTPServer | undefined
+  let occupant: net.Server | undefined
+
+  beforeEach(async () => {
+    mailDir = await mkdtemp(join(tmpdir(), 'maildev-start-err-'))
+    storage = new MemoryStorage()
+    await storage.initialize()
+  })
+
+  afterEach(async () => {
+    await server?.stop()
+    if (occupant) {
+      await new Promise<void>((resolve, reject) => {
+        occupant!.close((err) => (err ? reject(err) : resolve()))
+      })
+      occupant = undefined
+    }
+    await rm(mailDir, { recursive: true, force: true })
+  })
+
+  it('rejects start() with EADDRINUSE when the SMTP port is already bound', async () => {
+    occupant = net.createServer()
+    await new Promise<void>((resolve) => {
+      occupant!.listen(0, '127.0.0.1', () => resolve())
+    })
+    const port = (occupant.address() as AddressInfo).port
+
+    const uncaught: Error[] = []
+    const onUncaught = (err: Error) => {
+      uncaught.push(err)
+    }
+    process.on('uncaughtException', onUncaught)
+
+    try {
+      server = new SMTPServer({
+        storage,
+        mailDir,
+        port,
+        host: '127.0.0.1',
+        logger: false,
+      })
+      await expect(server.start()).rejects.toMatchObject({ code: 'EADDRINUSE' })
+      await new Promise<void>((resolve) => setImmediate(resolve))
+      expect(uncaught).toEqual([])
+    } finally {
+      process.removeListener('uncaughtException', onUncaught)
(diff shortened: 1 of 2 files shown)


djfarrelly:
Thanks for the contribution @cpruijsen - this looks good to me - We'll get this into the next release!

djfarrelly (approved):

PR #581, opened 2026-09-10
fix: honor MAILDEV_AUTO_RELAY and MAILDEV_AUTO_RELAY_RULES env vars (#327)

### LINKED ISSUE #327: MAILDEV_AUTO_RELAY environment variable ignored
I have `MAILDEV_AUTO_RELAY` set to no value using the docker beta 3 and auto relay is not enabled even though I have rules set. How do I enable auto relay without specifying an email address?

cpruijsen:
## Summary

`MAILDEV_AUTO_RELAY` and `MAILDEV_AUTO_RELAY_RULES` were documented (README) and used in Docker/Kubernetes, but `loadEnvConfig` never mapped them, so setting them (empty, `true`, or an email) had no effect. Auto-relay only worked via `--auto-relay [email]` or a config file.

`MAILDEV_AUTO_RELAY=true` was also the historical trap: when the value *was* read (v2 / custom images), it was treated as a recipient named `true`. Putting the var through the generic boolean parser would make an email like `qa@example.com` parse as `false`.

`loadEnvConfig` now maps:

- `MAILDEV_AUTO_RELAY=true`, `1`, or present-but-empty (`MAILDEV_AUTO_RELAY=`) → `autoRelay: true` (relay to each mail’s own recipients, honoring rules)
- any other `MAILDEV_AUTO_RELAY` value → that string as the override recipient
- `MAILDEV_AUTO_RELAY_RULES` → `autoRelayRules` (path)

Unset still means off. The config model already had `autoRelay?: boolean | string`; this is only the env wiring.

**What I chose:** special-case `boolean | string` in `env.ts`, matching `--auto-relay [email]`.
**Alternative:** treat the env var as boolean-only (`BOOLEAN_VARS`) and keep the recipient override on the CLI flag / config file. That would document `MAILDEV_AUTO_RELAY=true` and reject emails, but it would not match the README column or the maintainer note on this issue. Happy to switch if you would rather treat `false`/`0` as off (today those strings become the recipient, per “any other value is the override email”).

Fixes #327

## Test plan

- [x] `MAILDEV_AUTO_RELAY=true` / `1` / `''` → `autoRelay === true` (not `"true"`)
- [x] `MAILDEV_AUTO_RELAY=qa@example.com` → that string
- [x] `MAILDEV_AUTO_RELAY_RULES=/config/rules.json` → that path
- [x] vars absent → `autoRelay` / `autoRelayRules` unset
- [x] `pnpm --filter maildev test` (111 passed)
- [ ] Docker: `MAILDEV_AUTO_RELAY=` plus `MAILDEV_AUTO_RELAY_RULES` enables relay to rule matches without a catch-all address
- [ ] `MAILDEV_AUTO_RELAY=you@example.com` overrides recipients

### DIFF
--- a/packages/cli/src/config/env.ts
+++ b/packages/cli/src/config/env.ts
@@ -76,12 +76,16 @@ function parseBoolean(value: string): boolean {
 
 /**
  * MAILDEV_AUTO_RELAY is boolean | string: true/1/empty enable relay to
- * each mail's own recipients; any other value is the override recipient.
+ * each mail's own recipients; false/0 disable; any other value is the
+ * override recipient.
  * Do not put this in BOOLEAN_VARS — that would parse an email as false.
  */
 function parseAutoRelay(value: string): boolean | string {
   if (value === '' || value.toLowerCase() === 'true' || value === '1') {
     return true
+  }
+  if (value.toLowerCase() === 'false' || value === '0') {
+    return false
   }
   return value
 }
--- a/packages/cli/src/config/env.ts
+++ b/packages/cli/src/config/env.ts
@@ -32,6 +32,8 @@ const ENV_MAPPING: Record<string, keyof PartialMailDevConfig> = {
   MAILDEV_OUTGOING_PORT: 'outgoingPort',
   MAILDEV_OUTGOING_USER: 'outgoingUser',
   MAILDEV_OUTGOING_PASS: 'outgoingPass',
+  MAILDEV_AUTO_RELAY: 'autoRelay',
+  MAILDEV_AUTO_RELAY_RULES: 'autoRelayRules',
 
   // Storage
   MAILDEV_MAIL_DIRECTORY: 'mailDirectory',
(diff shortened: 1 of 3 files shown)


packages/cli/src/config/env.ts
```diff
@@ -72,6 +74,18 @@ function parseBoolean(value: string): boolean {
...
+ * MAILDEV_AUTO_RELAY is boolean | string: true/1/empty enable relay to
+ * each mail's own recipients; any other value is the override recipient.
+ * Do not put this in BOOLEAN_VARS — that would parse an email as false.
+ */
+function parseAutoRelay(value: string): boolean | string {
+  if (value === '' || value.toLowerCase() === 'true' || value === '1') {
+    return true
+  }
+  return value
+}
```
djfarrelly, line 87: > Happy to switch if you would rather treat false/0 as off (today those strings become the recipient, per “any other value is the override email”).

Thanks for calling this out in the PR description. I do think we should explicitly handle `'false'` (after toLowerCase) and `'0'` here for the reasons you mentioned. This would ensure that instead of the previous no-op, setting 0 or false would try to send an email to that address which would fail. 

I think adding a conditional and returning false would avoid the other issues.
```ts
if (value.toLowerCase() === 'false' || value === '0') {
  return false;
}
```

djfarrelly (changes requested):
Hey @cpruijsen - thanks for the contribution! I left one comment related to what you documented in your PR that I'd like to see addressed before merging. 

Also, can you add a changeset to this PR for our changelog via `pnpm changeset`?

Thanks - with those addressed, I'll get this into the next release!

cpruijsen:
Both addressed, thanks for the quick review.

`parseAutoRelay` now returns `false` for `false` (case-insensitive) and `0`, so those disable relay instead of being treated as an override recipient. Empty, `true` and `1` still enable it, and anything else is still the recipient address.

Changeset added as `.changeset/fix-auto-relay-env.md`.

djfarrelly (approved):
Thanks @cpruijsen - will get this merged and into the next release!

Issue #266, opened 2019-09-17
Embedded images using Docker


cropr:
I have an issue using maildev in a Docker container.  When embedding images the resulting source
looks like 
`<img src="//maildev-server/email/ccbt7Agm/attachment/phoneregistration.png">`
The hostname maildev-server is unknown and the image is not displayed.  
On non Docker maildev environments the hostname is localhost:1080 which works fine

djfarrelly:
Confirmed this is still a bug, and here's the root cause. When rendering an email's HTML, embedded (CID) image `src`s are rewritten to point at the attachment endpoint, and the base is taken from the request's **`Host` header**:

- `packages/api/src/server.ts:289` — `const baseUrl = request.headers.host ?? ''`
- `packages/smtp/src/sanitize.ts:92` — `const base = baseUrl ? \`//${baseUrl}\` : ''` → `src="//<host>/email/<id>/attachment/<file>"`

Behind Docker / a reverse proxy, the forwarded `Host` header isn't a URL the browser can reach (e.g. `maildev-server`), so you get `//maildev-server/email/.../attachment/...` and the image fails to load. Interestingly the code already produces the correct **host-independent relative** URL (`/email/...`) when `baseUrl` is empty.

Likely fix: drop the `//${baseUrl}` prefix and always emit the relative `/email/<id>/attachment/<file>` form (the browser resolves it against the page origin, which works in every deployment). Note `packages/smtp/src/__tests__/sanitize.test.ts` currently asserts the `//host` behavior, so that test would update too. Leaving open as a `bug`; PR welcome.

djfarrelly:
Will be fixed in next 3.0 release

Issue #439, opened 2022-12-08
Maildev behind apache virtual directory fails to load CID:Images


balajeeoshift:
The email content in the EML file is 

`<tr><td valign=3D"top" height=3D"52" width=3D"600" style=3D"padding:0px 24px 0px 24px;"><div style=3D"text-align:left;"><a href=3D"https://domain.com?utm_source=xxxxx" style=3D"text-decoration:none;"><span styl=
e=3D"height:51"><img src=3D"**cid:448997419133796281@E.G**" alt=3D"" width=3D"140" height=3D"51" border=3D"0" style=3D"display:block;margin-left:0;margin-right:auto;">
</span></a></div>
</td>
</tr>`

HTML content on the Maildev application is 

`<img src="//testdomain.com/email/J0t6L7Mh/attachment/logo.png" alt="" width="140" height="51" border="0" style="display:block;margin-left:0;margin-right:auto;">
`

I have the following apache virtual directory config for mail dev
    <Location /smtpdev/>
        RequestHeader set X-Request-URL "https://testdomain.com/smtpdev/"
        RequestHeader set X-Forwarded-Proto "https"
        RequestHeader set X-Forwarded-Port 443
        ProxyPreserveHost On
        ProxyPass  "http://127.0.0.1:2580/"
        ProxyPassReverse "http://127.0.0.1:2580/"
    </Location>

Version : MailDev 2.0.5

**Image missing in Email**
![image](https://user-images.githubusercontent.com/21282827/206507546-21b39742-55ba-42be-b155-45521d2078c0.png)

djfarrelly:
Confirming this is still a bug when MailDev is served under a subpath (e.g. behind apache/nginx with `--base-pathname /mail`): inline CID images don't load because the generated attachment URL doesn't include the base path.

Root cause — `buildAttachmentUrl` in `packages/smtp/src/sanitize.ts` builds the src purely from the request Host header and a fixed `/email/...` path:

```ts
const base = baseUrl ? `//${baseUrl}` : ''
return `${base}/email/${emailId}/attachment/${encodeURIComponent(filename)}`
```

The configured `basePathname` (used to prefix the actual routes and the UI mount in `orchestrator.ts` / `api/server.ts`) is never plumbed into this function, so behind `/mail` the img src comes out without the `/mail` prefix and 404s.

[...]

PR #582, opened 2026-09-13
Build CID attachment URLs against the API route and base path

dualfroz:
`replaceCidReferences` rewrote `cid:` links to `/email/:id/attachment/:file`, a route that does not exist, and it now uses `${basePath}/api/...`.
Note that the symptom in #439 no longer reproduces on main, since mailparser already inlines these parts as `data:` URIs.

Refs #439


packages/smtp/src/sanitize.ts
```diff
@@ -82,15 +84,18 @@ export function replaceCidReferences(
...
   emailId: string,
   filename: string,
-  baseUrl?: string
+  baseUrl?: string,
+  basePath?: string
 ): string {
   const base = baseUrl ? `//${baseUrl}` : ''
-  return `${base}/email/${emailId}/attachment/${encodeURIComponent(filename)}`
+  const prefix = basePath ? `/${basePath.replace(/^\/|\/$/g, '')}` : ''
+  return `${base}${prefix}/api/email/${emailId}/attachment/${encodeURIComponent(filename)}`
```
djfarrelly: For base path, while the CLI normalizes `"/"` to `""`, folks using packages programmatically may still pass `"/"` which would result in `//api/email` which would be invalid as a protocol relative URL. Can you add a case to handle when prefix resolves as `/` and omit from including it in the returned string?


One idea:

```suggestion
  const segments = basePath ? basePath.replace(/^\/+|\/+$/g, '') : ''
  const prefix = segments ? `/${segments}` : ''
  return `${base}${prefix}/api/email/${emailId}/attachment/${encodeURIComponent(filename)}`
```

packages/smtp/src/__tests__/sanitize.test.ts
```diff
@@ -96,7 +96,30 @@ describe('replaceCidReferences', () => {
...
+    const html = '<img src="cid:image123">'
+    const attachments = [
+      { contentId: 'image123', generatedFileName: 'abc.png' },
+    ]
+
+    for (const basePath of ['mail', '/mail', '/mail/']) {
+      const result = replaceCidReferences(html, 'email1', attachments, undefined, basePath)
+      expect(result).toContain('src="/mail/api/email/email1/attachment/abc.png"')
+    }
   })
```
djfarrelly, line 123: Two extra cases worth locking in while we're here (the first needs the sanitize.ts
fix, the second works already — I verified `//example.com/mail/api/...`
composes correctly):

```suggestion
  })

  it('should treat basePath "/" as no prefix', () => {
    const html = '<img src="cid:image123">'
    const attachments = [{ contentId: 'image123', generatedFileName: 'abc.png' }]
  
    const result = replaceCidReferences(html, 'email1', attachments, undefined, '/')
  
    expect(result).toContain('src="/api/email/email1/attachment/abc.png"')
  })
  
  it('should combine baseUrl and basePath', () => {
    const html = '<img src="cid:image123">'
    const attachments = [{ contentId: 'image123', generatedFileName: 'abc.png' }]
  
    const result = replaceCidReferences(html, 'email1', attachments, 'example.com', '/mail')
  
    expect(result).toContain('src="//example.com/mail/api/email/email1/attachment/abc.png"')
  })
```

djfarrelly (changes requested):
Hey @dualfroz - thanks for the contribution and fix! This looks _very_ close, but I noticed one thing that we should resolve before merging. I appreciate your work.

As a bonus, I think this should also close #266 too

dualfroz:
Hey. i kept your version of the prefix handling and added two tests.

djfarrelly (approved):

djfarrelly:
Thanks @dualfroz!

djfarrelly (approved):

